PR #8, opened 2018-09-19
Refactor the API

Kashomon:
Bunch of refactors for the API that have been brewing. Mostly renaming and removing cruft.

- Rename applications as components
- Add a version field for each component
- Use URLs for files
- Vastly reduce the number of fields in the Node Config (formerly ImageConfig)
- Rename Overlay to Patch.


Kashomon:
FYI: @mtaufen

PR #11, opened 2018-09-20
gcr.io/google-containers -> k8s.gcr.io

dekkagaijin:
Use the real repo, rather than rely on one of GCR's docker bug mitigations


dekkagaijin:
@kashomon ptal

ixdy:
why not just use the alias `k8s.gcr.io`?

dekkagaijin:
@ixdy that'd be fine, too

Kashomon:
LGTM

PR #40, opened 2018-10-14
Add a release channel field.

Kashomon:
This PR adds a channel field which is meant to indicate the stability of
a component, or in the case of the bundle, the minimum stability of all
the components in the bundle.

Open question: Maybe this should be an (opinionated) enum with Alpha,
Dev, Beta, Stable.


pkg/apis/bundle/v1alpha1/cluster_bundle.proto
```diff
@@ -53,21 +53,29 @@ message ClusterBundleSpec {
   // If a bundle is versioned, then all its components must be versioned.
   string version = 1;
 
+  // The release channel for this bundle, which indicates the minimum
+  // stability of components in the Bundle. For example, you might have
+  // 'Dev', 'Beta', or 'Stable' channels. If the channel is 'Stable' all
+  // artifacts in the bundle must be labeled as 'Stable'. If a channel is
```
johnsonj: Nit, I think. I'm reading this to say that the channel establishes a 'minimum stability' level of all packages.

I think it's reasonable to expect a Stable bundle to include Beta components that are deemed to play nicely with the Stable components. Would it be reasonable to drop this restriction? (or am I misinterpreting things?)

Kashomon:
From talking to @justinsb and @johnsonj -- Channels would be a concept outside of bundles. Like something you would subscribe to.

Issue #45, opened 2018-10-15
API: Component spec 


Kashomon:
Right now, the components don't have a spec / status. This make them problematic to apply to the cluster and act on them. I think this means that they probably need to be full spec / status objects (even though that adds a bit of noise to the schema)

Issue #42, opened 2018-10-14
API: Move Patching, NodeConfig to Extensions


Kashomon:
Here's my idea for unifying patching and node config, both of which have seemed pretty adhoc: Move them to an 'extensions' API and include them as just ordinary objects in a component.

`pkg/apis/extensions/v1alpha1/` for example.

This has the advantage of greatly simplifying what a component package is. Moreover, these objects can reuse the searching and externalization infrastructure already existing in components.

Issue #46, opened 2018-10-15
API: Status for Bundle and Components?


Kashomon:
Right now the Bundle is just a spec. The components will also be converted to be in a spec too. But that opens tho question: What should be in the Status? It seems like there should be something

Kashomon:
I don't think is necessary or useful for now, but we can revisit later

PR #49, opened 2018-10-16
Refactor the API to move NodeConfig, Patch to Extensions. (Breaking Change)

Kashomon:
This will hopefully be the last of the mega API refactors, and even this one is a bit smaller than previous.

- Add a Spec to ComponentPackage. This allows us flexibility later if we decide we want a `status` field.
- Move NodeConfig to be a Bundle Extension rather than a core part of the Bundle. Now everything's just components!
- Re-add the Patch (as PatchTemplate) to Bundle Extensions. It's not currently used, but it's a good example of something that could be a bundle extenison.

Addresses: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/46
Fixes: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/45
Fixes: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/42


jplobosgle:
LGTM

PR #50, opened 2018-10-16
Fix nil assertion

Kashomon:
Remove an erroneous nil check and add better messaging for the previous nil
check.


jplobosgle:
LGTM

PR #52, opened 2018-10-17
Fix a bug in the inline process.

Kashomon:
Fixes a bug during inlining raw-text files: the files weren't getting
cleared. Add a test that ensures the files are cleared.


jplobosgle:
LGTM

Issue #57, opened 2018-10-18
Inlining raw-text doesn't work for nested child components


Kashomon:
Looks like inlining raw-text doesn't work for nested child components. Basically, just need to replicate the following for raw text also:

```go
» » » for _, ocf := range comp.GetSpec().GetClusterObjectFiles() {
» » » » compUrl := cf.GetUrl()
» » » » objUrl := ocf.GetUrl()
» » » » if strings.HasPrefix(objUrl, "file://") && strings.HasPrefix(compUrl, "file://") {
» » » » » ocf.Url = "file://" + filepath.Join(filepath.Dir(shortFileUrl(compUrl)), shortFileUrl(objUrl))
» » » » }
» » » }
```

PR #58, opened 2018-10-18
Fix inlining for raw text.

Kashomon:
Fixes: #57

### DIFF
--- a/pkg/transformer/inline.go
+++ b/pkg/transformer/inline.go
@@ -140,9 +140,17 @@ func (n *Inliner) processComponentPackageFiles(ctx context.Context, b *bpb.Clust
 					cf.GetUrl())
 			}
 
+			compUrl := cf.GetUrl()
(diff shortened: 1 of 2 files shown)


pkg/transformer/inline.go
```diff
@@ -148,6 +148,15 @@ func (n *Inliner) processComponentPackageFiles(ctx context.Context, b *bpb.Clust
 					ocf.Url = "file://" + filepath.Join(filepath.Dir(shortFileUrl(compUrl)), shortFileUrl(objUrl))
 				}
 			}
+
+			// Do the same with the text files.
+			for _, ocf := range comp.GetSpec().GetRawTextFiles() {
+				compUrl := cf.GetUrl()
```
jplobosgle: Nit: 'compUrl' is the same every iteration, so it can be pulled out of this for loop, and from the loop on line 145, to line 143.

Kashomon: Done.

Issue #63, opened 2018-10-20
Validation: bundle kinds and apiVersion


Kashomon:
Currently the validation doesn't validate the apiVersion or kind fields of the bundle or component object, and it definitely should.

PR #67, opened 2018-10-23
Add validation for kind / apiversion.

Kashomon:
This PR adds validation for the kind and API Version for the bundle and
the cluster components.

Fixes: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/63

### DIFF
--- a/pkg/validation/validate_bundle.go
+++ b/pkg/validation/validate_bundle.go
@@ -16,6 +16,7 @@ package validation
 
 import (
 	"fmt"
+	"regexp"
 
 	bpb "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/apis/bundle/v1alpha1"
 	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/core"
@@ -26,6 +27,8 @@ type BundleValidator struct {
 	Bundle *bpb.ClusterBundle
 }
 
+var apiVersionPattern = regexp.MustCompile(`^gke.io/k8s-cluster-bundle/\w+$`)
+
 // NewBundleValidator creates a new Bundle Validator
 func NewBundleValidator(b *bpb.ClusterBundle) *BundleValidator {
 	return &BundleValidator{b}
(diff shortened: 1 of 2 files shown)


pkg/validation/validate_bundle.go
```diff
@@ -34,11 +35,29 @@ func NewBundleValidator(b *bpb.ClusterBundle) *BundleValidator {
...
 }
 
+func (b *BundleValidator) validateBundle() []error {
+	var errs []error
+	n := b.Bundle.GetMetadata().GetName()
+	if n == "" {
+		errs = append(errs, fmt.Errorf("bundle name was empty, but must always be present"))
+	}
+	api := b.Bundle.GetApiVersion()
+	if !strings.Contains(api, "gke.io/k8s-cluster-bundle/") {
```
jplobosgle: The error message (and the comment for the bundle, https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/apis/bundle/v1alpha1/cluster_bundle.proto#L28) say that the format is 

"gke.io/k8s-cluster-bundle/<version>"

So, 'Contains' will match strings like "foogke.io/k8s-cluster-bundle/bar" or "gke.io/k8s-cluster-bundle/" (which doesn't have the version part)

So, perhaps something like this: 
ver := "gke.io/k8s-cluster-bundle/"
if !(strings.Index(api, ver) == 0 && len(api) > len(ver))  {
   ....
}

Or, with a regex (https://play.golang.org/p/v17peCDPOg7): 
matched, err := regexp.MatchString("^gke\\.io\\/k8s\\-cluster-bundle\\/[a-zA-Z0-9]+$", api) 
if !matched || err != nil {
   ....
}

(regex doesn't enforce v1alpha1/v1beta1/v1, etc, just alphanumeric versions)

Kashomon: Good point; I got lazy =)

Kashomon: Done

PR #68, opened 2018-10-25
Add a Go IDL kubernetes type for the bundle and components.

Kashomon:
I mostly finished client generation and CRD generation, but hit a
blocking issue in deepcopy-gen, which blocks registry-gen and thus client-gen: https://github.com/kubernetes/gengo/issues/128

So, I've added CRD generation + some sed hacks to make it work. And I've
added the other generation command in hack/

Also note: the group for the CRD does not match the group that I'm currently using. In the bundle, I have been using `gke.io/k8s-cluster-bundle`. The CRD generator would really rather I use `bundle.gke.io`. I'm not sure which is right, so I've left this discrepancy for discussion. I personally like `bundle.gke.io` slightly better.


pkg/apis/k8sbundle/v1alpha1/bundle_types.go
```diff
@@ -0,0 +1,141 @@
...
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UnstructuredJSON represents JSON not bound to any particular schema
+type UnstructuredJSON map[string]interface{}
```
lavalamp: We have RawExtension for this. You can't use interface{} in an api type.

pkg/apis/k8sbundle/v1alpha1/bundle_types.go
```diff
@@ -0,0 +1,141 @@
...
+	// A list of components that must be packaged in a bundle in with this
+	// component.
+	Requirements []MinRequirement `json:"requirements,omitempty"`
+
+	// Structured Kubenetes objects that run as part of this app, whether on the
+	// master, on the nodes, or in some other fashio.  These Kubernetes objects
+	// are inlined and must be YAML/JSON compatible. Each must have `apiVersion`,
+	// `kind`, and `metadata`.
+	//
+	// This is essentially equivalent to the Kubernetes `Unstructured` type.
```
lavalamp: It's RawExtension in API types.

lavalamp: If you want to force yaml/json encoding, why not take a list of strings?

pkg/apis/k8sbundle/v1alpha1/bundle_types.go
Kashomon: Pardon the ignorance, but is there a technical reason why you can't use interface{}?

pkg/apis/k8sbundle/v1alpha1/bundle_types.go
Kashomon: I've considered it. Then you'd rely on validation, which is certainly possible.

There's a lot of power in having objects forced into a parsed JSON-structure. But it sounds maybe that's not possible in k8S/

lavalamp: The thing is, if you don't want people to have to put in quoted yaml/json, you have to do a bunch of tricks during unmarshalling to make it work. quoted json in json is a huge pain, although it's not too bad in yaml.

Anyway, the RawExtension object does this for you.

pkg/apis/k8sbundle/v1alpha1/bundle_types.go
lavalamp: Well, the tooling doesn't support it, for one :)

But I think that's more or less a deliberate choice?
* there's no way to document for users what they're actually communicating to the server
* there's no way to promise things that process API types that they're guaranteed to only get safe things inside the interface{}'s (type safety)

pkg/apis/k8sbundle/v1alpha1/bundle_types.go
Kashomon: Note to self: https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/types.go#L92

So maybe that's the story.

Kashomon: That was one of the beautiful things about using components that are inlined as JSON-format is that it avoids all the quoting nonsense and also is really nice for readability. I'll think about using the RuntimeExtension.

pkg/apis/k8sbundle/v1alpha1/bundle_types.go
Kashomon: Isn't the same also true for RawExtension?

Kashomon: In any case, I think I would add this in a separate PR and ensure that proto and this file are in-sync. I'm warming to the RawExtension type and I like the idea of compatibility with the broader K8S stuff.

Kashomon:
So ignoring the RawExtension discussion for now (which needs a follow-up, I'll file a new issue). Do others thing this is reasonable to go in?

hack/update-codegen.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+command -v deepcopy-gen >/dev/null 2>&1 || {
```
justinsb: FYI it's probably easier to vendor it and then run it from the vendored directory :-)

justinsb:
I think merging this in and then doing the RawExtension change separately would actually probably be good in terms of isolating the changes.

/approve
/lgtm

hack/update-codegen.sh
Kashomon: I absolutely agree in terms of UX. The concern here is that the code-generator tools add extra dependencies that I don't think I need. But maybe there's a way to say: hey I need this for tooling / tests but you don't need to pull this in as a consumer of this library?

Kashomon:
RawExtension follow-up @ #69

PR #71, opened 2018-10-29
Issue 25/26: Enforce SemVer string format and validate MinRequirements.

jplobosgle:


pkg/validation/validate_bundle.go
```diff
@@ -27,7 +27,21 @@ type BundleValidator struct {
...
+	apiVersionPattern = regexp.MustCompile(`^gke.io/k8s-cluster-bundle/\w+$`)
+
+	// Regex string for numbers without leading zeros.
+	number = `([1-9]\d*|0)`
+	// Regex string for dot separated alphanumeric identifier with hyphen.
+	dotID  = `[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*`
+	// The three regex strings below make up the SemVer 2.0.0 string.
+	// MAJOR.MINOR.PATCH-preRelease+metaData
+	majorMinorPatch = fmt.Sprintf(`%s\.%s\.%s`, number, number, number)
+	preRelease      = fmt.Sprintf(`(-%v)?`, dotID)
```
Kashomon: Same comment from previous: I think we probably should just use Major.Minor.Patch for the version of the bundle. I.e., Let's start out more restrictive and go to less restrictive. WDYT?

Kashomon: It's possible the component API version should have less restrictive versioning-matching (i.e., the full semVerPattern). But hopefully we won't need that?

Kashomon:
Note: in #77, I'm removing min requirements from the API

PR #75, opened 2018-10-31
Switch to unstructured.Unstructured type and finish Client-gen

### LINKED ISSUE #69: RawExtension for arbitrary JSON
Kubernetes uses the RawExtension type to represent arbitrary JSON components. Right now in the proto-world, I'm using `Struct` and in the Go IDL version (soon to come) I'll probably use `map[string]interface{}` for compatibility. But this is non-standard and the standard type for this purpose is RawExtension.

https://github.com/kubernetes/apimachinery/blob/ed135c5b96450fd24e5e981c708114fbbd950697/pkg/runtime/types.go

```go
type RawExtension struct {
	// Raw is the underlying serialization of this object.
	//
	// TODO: Determine how to detect ContentType and ContentEncoding of 'Raw' data.
	Raw []byte `protobuf:"bytes,1,opt,name=raw"`
	// Object can hold a representation of this extension - useful for working with versioned
	// structs.
	Object Object `json:"-"`
}
```

### LINKED ISSUE #74: Add a CRD + Generated client for Bundles / Components
We should add a CRD + Generated client for bundles and components to make them easier to use.

Kashomon:
This PR switches to Unstructured for representing unknown K8S objects,
which allows us to finish client generation.

This best corresponds to the contents and is compatible with the current
Proto schema and I think is a good semantic fit. We can't know the
schema of all the types, just that they have K8S-shape.

Fixes #69 
Fixes #74


pkg/apis/bundle/v1alpha1/component_package.proto
```diff
@@ -93,6 +93,12 @@ message ComponentPackageSpec {
   repeated File raw_text_files = 6;
 }
 
+// Unstructured represents raw Kubernetes objects and corresponds to the
+// Kubernetes `unstructured.Unstructured` type from apimachinery.
+message Unstructured {
+  google.protobuf.Struct zbject = 1;
```
johnbelamaric: zbject is a typo?

Kashomon: Oops. that was for testing but I forgot to remove.

Issue #76, opened 2018-10-31
Rethink minRequirements design


Kashomon:
For now, the min requirements and component API version will be removed. This sort of version dependency is hard and needs a more careful design.

PR #77, opened 2018-10-31
[Breaking] Remove the min-requirements and component api version fields.

Kashomon:
This PR removes the min requirements and component api version fields. 

I think we still want / need something like this, but I'd like to be a little more intentional about what this field looks like. #76 tracks re-thinking this feature.


Kashomon:
(just rebase-fixing-changes). Going to merge.

PR #78, opened 2018-10-31
[Breaking] Add name as an explicit field

Kashomon:
This PR adds 'name' as an explicit field. It's becoming pretty clear that having an explicit name for the component is useful. For example, we might have more than one version of a component running around in a cluster. And if that's the case, then Metadata.Name is no longer a good fit: that field must be unique (at least within a particular namespace).

An alternative we played with was using labels or annotations for this field. But in a real sense, the key for the component is (name, tuple), so it didn't seem like a great fit. In the end, having this well defined and clear makes it easier to build automation.


Kashomon:
Oops. this is the wrong pr.

PR #79, opened 2018-10-31
[Breaking] Add explicit bundle and component name.

Kashomon:
This PR adds 'name' as an explicit field. It's becoming pretty clear that having an explicit name for the component is useful. For example, we might have more than one version of a component running around in a cluster. And if that's the case, then Metadata.Name is no longer a good fit: that field must be unique (at least within a particular namespace).

An alternative we played with was using labels or annotations for this field. But in a real sense, the key for the component is (name, tuple), so it didn't seem like a great fit. In the end, having this well defined and clear makes it easier to build automation.


Kashomon:
Note: this is rebased on the Min-requirements removal PR.

pkg/apis/bundle/v1alpha1/cluster_bundle.proto
```diff
@@ -42,6 +41,9 @@ message ClusterBundle {
 
 // A spec object that wraps the Cluster Bundle.
 message ClusterBundleSpec {
+  // Name of the cluster bundle.
+  string name = 1;
```
mtaufen: Is my understanding that the "name" string here is intended to identify a collection of bundles, rather than a specific bundle?

I'm worried that overloading "name" is going to be confusing, because most will assume ObjectMeta.Name when "name" is used in conversation - and in that context "name" always maps to a singular object per namespace per kind. 

Some possible other options:
- class
  - "Class" in K8s is often used to describe classes of resources that can be scheduled: StorageClass, ResourceClass, RuntimeClass.
  -  Since Bundles could plausibly represent the flavors of clusters one is allowed to create (and maybe schedule to), maybe Bundle<=>ClusterClass is true if you squint at it. But I kind of had to squint to make this one work.
- category (casually used in K8s... shows up a near taints a lot when you grep for "category", but not a very common term)
- type (most common K8s use seems to be for sum types)
- kind (this *is* the word that K8s uses to describe config types. but like "name" it is has a specific, common meaning, and could be confusing in conversation)

Aaaargh none of these seem very good!

pkg/apis/bundle/v1alpha1/cluster_bundle.proto
```diff
@@ -50,15 +52,15 @@ message ClusterBundleSpec {
...
   //
   // If a bundle is versioned, then all its components must be versioned.
-  string version = 1;
+  string version = 2;
 
   // Kubernetes objects grouped into component packages and versioned together.
   // These could be applications or they could be some sort of supporting
   // object collection.
-  repeated ComponentPackage components = 2;
+  repeated ComponentPackage components = 3;
```
mtaufen: Isn't our plan to make this a reference?

Kashomon: Yes, but this is a massive change; and might not make Alpha 1.X given the size.

Kashomon: I started doing it in this PR, but then reverted it.

johnbelamaric: Yes, we discussed this being a name and version that reference a separate component CRD.

Kashomon: Let me be more clear: this is definitely coming, but I wanted to incrementally approach that.

pkg/apis/bundle/v1alpha1/cluster_bundle.proto
johnbelamaric: Ok, so we can differentiate that the metadata.name is the resource name in a particular k8s api server that is really a reference back to some canonical "bundle" that is identified by this 'name' and a version. Maybe just "bundle name"?

mtaufen: I think a description of the concrete use-case where this is necessary would help this discussion.

Kashomon:
Sounds like the Spec.Name is significantly more controversial than for the component. I've moved the component to #81

PR #81, opened 2018-11-01
[Breaking] Add a Name field to the component spec

Kashomon:
This adds a new, explicit name field to the component spec, which represents the canonical 'name' of the component.


johnbelamaric:
Maybe we should call it cname or canonicalName?

Kashomon:
I like `canonicalName` for the component. `family` might be more appropriate for the component set.

johnbelamaric:
appName?

Issue #97, opened 2018-11-25
Validation: Add URL Validation


Kashomon:
Now that we parse File URLs as real urls, we should validate that all URLs are well formed.

Depends on https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/pull/96

Kashomon:
Fixed now. thanks @u5surf!

PR #100, opened 2018-11-27
Add URL validation #97

u5surf:
#97 
I added url validation following rules of Go’s net/url Parse.

### DIFF
--- a/pkg/validation/validate_url_test.go
+++ b/pkg/validation/validate_url_test.go
@@ -64,15 +64,6 @@ func TestValidateURL(t *testing.T) {
 					return
 				}
 				return // no error, not a problem
-			} else {
-				if tc.errSubstr == "" {
-					t.Errorf("got error %q, but didn't expect one", err.Error())
-					return
-				} else if !strings.Contains(err.Error(), tc.errSubstr) {
-					t.Errorf("got error %q, but expected it to contain %q", err.Error(), tc.errSubstr)
-					return
-				}
-				return // error matches
 			}
 		})
 	}
--- a/pkg/validation/validate_url_test.go
+++ b/pkg/validation/validate_url_test.go
@@ -23,21 +23,15 @@ func TestValidateURL(t *testing.T) {
 	testCases := []struct {
 		desc      string
 		in        string
-		scheme    string
-		path      string
 		errSubstr string
 	}{
 		{
 			desc:   "success: normal file url",
 			in:     "file:///foo/bar",
-			scheme: "file",
-			path:   "/foo/bar",
 		},
 		{
 			desc:   "success: normal file url with localhost",
 			in:     "file://localhost/foo/bar",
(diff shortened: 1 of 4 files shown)


Kashomon:
Thanks for the PR!

pkg/validation/validate_url_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			err := validateURL(tc.in)
+			if err == nil {
+				if tc.errSubstr != "" && !strings.Contains(err.Error(), tc.errSubstr) {
+					t.Errorf("got nil error, but expected one with %q", tc.errSubstr)
+					return
+				}
+				return // no error, not a problem
+			} else {
```
Kashomon: nit: don't need the else here.

Kashomon:
Can you update the bazel BUILD file? Once you have `bazel` [installed](https://docs.bazel.build/versions/master/install.html), you can run `bazel run //:gazelle`. Or, I suppose you can just run `gazelle` manually: https://github.com/bazelbuild/bazel-gazelle#running-gazelle-with-go

u5surf:
> Can you update the bazel BUILD file? Once you have `bazel` [installed](https://docs.bazel.build/versions/master/install.html), you can run `bazel run //:gazelle`. Or, I suppose you can just run `gazelle` manually: https://github.com/bazelbuild/bazel-gazelle#running-gazelle-with-go

```
k8s-cluster-bundle $ bazel run //:gazelle
INFO: Analysed target //:gazelle (38 packages loaded, 7008 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle.bash
INFO: Elapsed time: 108.618s, Critical Path: 64.03s, Remote (0.00% of the time): [queue: 0.00%, setup: 0.00%, process: 0.00%]
INFO: 37 processes: 37 processwrapper-sandbox.
INFO: Build completed successfully, 50 total actions
INFO: Build completed successfully, 50 total actions
```

pkg/validation/validate_url_test.go
```diff
@@ -0,0 +1,70 @@
...
+			desc: "success: no scheme",
+			in:   "foo/bar/biff",
+			path: "foo/bar/biff",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			err := validateURL(tc.in)
+			if err == nil {
+				if tc.errSubstr != "" && !strings.Contains(err.Error(), tc.errSubstr) {
```
Kashomon: this isn't quite right. err == nil means that err.Error() is dereferencing a nil object.

There should be a case where err != nil and you check this. I got this wrong in validate_names, which is probably why this is wrong here. I'll go fix that.

Issue #102, opened 2018-11-28
RawTextMaker: Return all objects


Kashomon:
The raw text maker currently just returns raw text. It should 'make' using the raw text objects, but it should return all objects.

PR #110, opened 2018-11-29
Rework the maker package as an options package

Kashomon:
This PR works to clean up a bunch of stuff from the previous mega PR.

- Rename maker to options.
- Remove the filter option. There's no reason to have that in the API.
  It's easy enough to do it earlier or later.
- Fix the raw text maker to return all the objects, but only
  de-templatize the raw-text one.

Smaller changes:
- Remove `deps/` directory
- Address some style comments on `helpers.go`

Fixes: #102


deps/README.md
```diff
@@ -1,3 +0,0 @@
-# Deps
```
ihmccreery: Why did this get deleted?

pkg/options/options.go
```diff
@@ -12,26 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package maker provides functionality for adding options to components
-package maker
+// Package options provides functionality for adding options to components.
```
ihmccreery: Please add more context here.  What is an option?  Why?  Examples?

pkg/options/rawyamltmpl/raw_text_maker.go
```diff
@@ -28,25 +28,23 @@ import (
 	bundle "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/apis/bundle/v1alpha1"
 	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/converter"
 	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/filter"
-	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/maker"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/options"
 )
 
-// Maker makes components via go-templating, for objects stored as raw strings
+// Applier applies options via go-templating, for objects stored as raw strings
```
ihmccreery: What ever happened to using Kinflate or whatever it's called now?

pkg/options/rawyamltmpl/raw_text_maker.go
```diff
@@ -28,25 +28,23 @@ import (
...
 // in config maps.
-type Maker struct{}
+type Applier struct{}
```
ihmccreery: File filename as well, here & below.

pkg/options/options.go
maisem: +1

pkg/options/simpletemplate/simple_maker.go
```diff
@@ -26,24 +26,21 @@ import (
...
 	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/converter"
-	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/filter"
-	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/maker"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/options"
 )
 
-// Maker makes components via go-templating.
-type Maker struct{}
+// Applier applies options via go-templating.
+type Applier struct{}
```
maisem: why is this public and not private with a public constructor?

deps/README.md
Kashomon: It was an empty directory

pkg/options/simpletemplate/simple_maker.go
Kashomon: Sure could. I don't have a strong opinion

pkg/options/rawyamltmpl/raw_text_maker.go
Kashomon: Kustomize (formerly Kinflate) could be another options applier, for sure. I haven't added it yet and might not for a little while. Just using go templates should get us some mileage before trying to add new customization systems.

pkg/options/rawyamltmpl/raw_text_maker.go
Kashomon: ah, good catch

pkg/options/options.go
Kashomon: Added some context

pkg/options/rawyamltmpl/raw_text_applier.go
```diff
@@ -28,17 +28,20 @@ import (
...
-// The raw text maker's MakeComponent method only uses inlined raw text as
-// input to the object-creation process.
-func (m *Maker) MakeComponent(comp *bundle.ComponentPackage, pm maker.ParamMaker, of *filter.Options) (*bundle.ComponentPackage, error) {
+// NewApplier creates a new optionts applier instance.
+func NewApplier() options.Applier {
+	return &applier{}
+}
+
+// ApplyOptions applies options to to raw-text that's been inlined.
+func (m *applier) ApplyOptions(comp *bundle.ComponentPackage, pm options.OptionsFn) (*bundle.ComponentPackage, error) {
```
ihmccreery: Use `common.ApplyCommon` (`common.ApplyOptions`, if you take my comment above)?

pkg/options/rawyamltmpl/raw_text_applier.go
```diff
@@ -28,17 +28,20 @@ import (
...
-// MakeComponent makes a component, implementing the make-component interface.
-// The raw text maker's MakeComponent method only uses inlined raw text as
-// input to the object-creation process.
-func (m *Maker) MakeComponent(comp *bundle.ComponentPackage, pm maker.ParamMaker, of *filter.Options) (*bundle.ComponentPackage, error) {
+// NewApplier creates a new optionts applier instance.
+func NewApplier() options.Applier {
+	return &applier{}
+}
+
+// ApplyOptions applies options to to raw-text that's been inlined.
```
ihmccreery: "to to" -> "to"

pkg/options/common.go
```diff
@@ -12,37 +12,34 @@
...
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	bundle "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/apis/bundle/v1alpha1"
-	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/filter"
 )
 
-// ObjHandler is a function that can 'make' objects.
-type ObjHandler func(obj *unstructured.Unstructured, ref bundle.ComponentReference, pm ParamMaker) (*unstructured.Unstructured, error)
+// ObjHandler is a function that can apply options to a specific Kubernetes
```
ihmccreery: Not specific, right?  It'll take any K8s object.

pkg/options/rawyamltmpl/raw_text_applier.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package rawtexttmpl is a special-case component maker for constructing
+// Package rawtexttmpl is a special-case option applier for constructing
 // objects with the assumption that objects have been inlined via RawText into
 // ConfigMaps.
```
ihmccreery: Why doesn't `SimpleApplier` work here?  I'd expect it to handle ConfigMaps effectively.

pkg/options/common.go
```diff
@@ -12,37 +12,34 @@
...
-// only the object handling logic.
-func MakeCommon(comp *bundle.ComponentPackage, p ParamMaker, of *filter.Options, fn ObjHandler) (*bundle.ComponentPackage, error) {
+// ApplyCommon provides common functionality for applying options, deferring
+// the specific object handling logic.
+func ApplyCommon(comp *bundle.ComponentPackage, optFn OptionsFn, objFn ObjHandler) (*bundle.ComponentPackage, error) {
 	comp = comp.DeepCopy()
 	ref := comp.ComponentReference()
 
 	if len(comp.Spec.Objects) == 0 {
 		return nil, fmt.Errorf("no objects found for component %v", ref)
```
ihmccreery: If this is really an error, please add tests for this behavior.  (It's not immediately clear this should be an error.)

pkg/options/rawyamltmpl/raw_text_applier.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package rawtexttmpl is a special-case component maker for constructing
+// Package rawtexttmpl is a special-case option applier for constructing
```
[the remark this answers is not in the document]
Kashomon: It means that these values were originally inlined as `RawTextFiles`. It's a little goofy, but it's going to be veryuseful when there are legacy templates that have heavy templating and so can't be parsed as YAML.

pkg/options/options.go
```diff
@@ -0,0 +1,45 @@
...
+//     - A map[string]interface{}, where interface{} has eth same properties as above.
+type JSONOptions map[string]interface{}
+
+// OptionsFn is a function that can create an options object.
+type OptionsFn func() (JSONOptions, error)
+
+// OptionsApplier represents an object that can take options and apply them to components.
+type Applier interface {
+	// ApplyOptions applys options to some subset objects from the component.
+	ApplyOptions(comp *bundle.ComponentPackage, p OptionsFn) (*bundle.ComponentPackage, error)
```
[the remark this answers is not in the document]
Kashomon: Oh, that seems fine.

pkg/options/rawyamltmpl/raw_text_applier.go
Kashomon: The raw text converter filters based on the inline annotation and then returns the sub-objects (rather just returning the reifed config maps).

pkg/options/rawyamltmpl/raw_text_applier.go
Kashomon: It can't because of the way the raw text templates work -- they're filegroups.

pkg/options/simpletemplate/simple_applier_test.go
```diff
@@ -45,7 +44,7 @@ func TestSimpleMaker(t *testing.T) {
 		t.Fatalf("Error converting component to yaml: %v", err)
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/options/simpletemplate/simple_applier_test.go
```diff
@@ -54,8 +53,7 @@ func TestSimpleMaker(t *testing.T) {
 
 	params, _ := pm()
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/options/simpletemplate/simple_applier_test.go
```diff
@@ -54,8 +53,7 @@ func TestSimpleMaker(t *testing.T) {
...
 
-	m := &Maker{}
-	made, err := m.MakeComponent(comp, pm, nil)
+	made, err := NewApplier().ApplyOptions(comp, pm)
 	if err != nil {
 		t.Fatalf("Error converting making component: %v", err)
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/options/simpletemplate/simple_applier_test.go
```diff
@@ -54,8 +53,7 @@ func TestSimpleMaker(t *testing.T) {
 
 	params, _ := pm()
 
-	m := &Maker{}
-	made, err := m.MakeComponent(comp, pm, nil)
+	made, err := NewApplier().ApplyOptions(comp, pm)
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/options/simpletemplate/simple_applier_test.go
```diff
@@ -54,8 +53,7 @@ func TestSimpleMaker(t *testing.T) {
...
 	if err != nil {
 		t.Fatalf("Error converting making component: %v", err)
 	}
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/options/common.go
Kashomon: Done

pkg/options/common.go
```diff
@@ -12,37 +12,34 @@
...
 
 	bundle "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/apis/bundle/v1alpha1"
-	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/filter"
 )
 
-// ObjHandler is a function that can 'make' objects.
-type ObjHandler func(obj *unstructured.Unstructured, ref bundle.ComponentReference, pm ParamMaker) (*unstructured.Unstructured, error)
+// ObjHandler is a function that can apply options to a specific Kubernetes
+// object.
+type ObjHandler func(obj *unstructured.Unstructured, ref bundle.ComponentReference, pm OptionsFn) (*unstructured.Unstructured, error)
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/options/common.go
```diff
@@ -12,37 +12,34 @@
...
 
-// MakeCommon provides common functionality for making components, deferring
-// only the object handling logic.
-func MakeCommon(comp *bundle.ComponentPackage, p ParamMaker, of *filter.Options, fn ObjHandler) (*bundle.ComponentPackage, error) {
+// ApplyCommon provides common functionality for applying options, deferring
+// the specific object handling logic.
+func ApplyCommon(comp *bundle.ComponentPackage, optFn OptionsFn, objFn ObjHandler) (*bundle.ComponentPackage, error) {
```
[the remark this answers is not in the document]
Kashomon: That was the intent. But the raw text package can't work that way because one ConfigMap can contain multiple objects. So, I think having this helper is maybe a bit more workable in the general case

pkg/options/options.go
```diff
@@ -0,0 +1,45 @@
...
+	bundle "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/apis/bundle/v1alpha1"
+)
+
+// TODO(kashomon): Should options be of JSON Type or just interface{}?
+
+// JSONOptions is an instance of options, represented as a JSON object
+// encoded as map[string]interface{} where each interface{} must be one of:
+//     - A primitive type
+//     - A slice of interface{}, where interface{} has the same properties as above.
+//     - A map[string]interface{}, where interface{} has eth same properties as above.
```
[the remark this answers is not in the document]
Kashomon: No, it's not. added a comment

pkg/options/options.go
```diff
@@ -0,0 +1,45 @@
...
+type JSONOptions map[string]interface{}
+
+// OptionsFn is a function that can create an options object.
+type OptionsFn func() (JSONOptions, error)
+
+// OptionsApplier represents an object that can take options and apply them to components.
+type Applier interface {
+	// ApplyOptions applys options to some subset objects from the component.
```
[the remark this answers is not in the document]
Kashomon: Done.

pkg/options/common.go
Kashomon: good point. It doesn't really need to be an error

pkg/options/rawyamltmpl/raw_text_applier.go
Kashomon: Done

pkg/options/rawyamltmpl/raw_text_applier.go
Kashomon: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/apis/bundle/v1alpha1/component_types.go#L117

has some more details

pkg/options/common.go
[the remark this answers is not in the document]
ihmccreery: But you're not using `ApplyCommon` in `RawTextApplier`, afaict.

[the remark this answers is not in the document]
Kashomon: Yeah, that's correct. I created the Common thing because i was thinking about the PatchTemplateApplier that I'm thinking about. Maybe this is over-design, but it's not too-far off I think.

pkg/options/rawyamltmpl/raw_text_applier.go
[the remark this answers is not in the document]
ihmccreery: So you're basically going to use a `ConfigMap` as a wrapper around arbitrary text, which will then get go-templated into first-class K8s objects, which will then ... get pulled out of the `ConfigMap`s and stored first-class?

Please add more context in comments.

pkg/options/rawyamltmpl/raw_text_applier.go
ihmccreery: Please add inline comments to that affect, with more details about why you can't use `common.ApplyCommon`.

pkg/options/rawyamltmpl/raw_text_applier.go
ihmccreery: Ooohhhh that clarifies this substantially.  Your comment above,

> The returned component should be copy of the original, with (perhaps) modifications made to the original.

seems to directly contradict this, though.  I think these concerns ought to be separated (ConfigMap containing templated objects -> ConfigMap containing complete objects -> complete objects).

Kashomon: Added some comments. Does this help?

pkg/options/rawyamltmpl/raw_text_applier.go
Kashomon: Added some comments

pkg/options/rawyamltmpl/raw_text_applier.go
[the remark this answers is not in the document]
Kashomon: Added some comments on the method.

pkg/options/rawyamltmpl/raw_text_applier.go
ihmccreery: Way better, yes, thanks.

pkg/options/rawyamltmpl/raw_text_applier.go
```diff
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package rawtexttmpl is a special-case component maker for constructing
-// objects with the assumption that objects have been inlined via RawText into
-// ConfigMaps.
+// Package rawtexttmpl is a special-case option applier for adding options to
+// objects with the assumption that objects have been inlined as go-templates
+// via RawTextFiles ComponentPackageSpec field into ConfigMaps.
```
ihmccreery: I think it's worth adding the 'Once the parameters are applied via the go template, ...' description to the package description as well, as a reader might very well look here first.

Kashomon: Done

Issue #89, opened 2018-11-15
Investigate using the stdlib for validation


Kashomon:
Core kubernetes has a bunch of validation built in. For example: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/validation/objectmeta.go

We should consider deferring to that where possible.

PR #111, opened 2018-11-30
Rely on K8S validation

Kashomon:
This reworks the validation library to rely on the OSS K8S machinery
validation logic.

Fixes: #89


pkg/validation/validate_component.go
```diff
@@ -47,75 +50,108 @@ func NewComponentValidator(c []*bundle.ComponentPackage, set *bundle.ComponentSe
...
-			errs = append(errs, fmt.Errorf("the component name %q was invalid config %v", n, ca))
+		if n == "" {
+			errs = append(errs, field.Required(cpathIdx(i).Child("Spec", "ComponentName"), "componentName is required"))
+		}
+		ver := ca.Spec.Version
+		if ver == "" {
+			errs = append(errs, field.Required(cpathIdx(i).Child("Spec", "Version"), "version is required"))
+		}
+		if n == "" || ver == "" {
+			// Other validation relies on components having a unique name+version pair.
```
mtaufen: s/Other/Subsequent
Is uniqueness of (n, ver) across the components a requirement of below validation? If not could just say "non-empty" in this comment.

pkg/validation/validate_component.go
```diff
@@ -47,75 +50,108 @@ func NewComponentValidator(c []*bundle.ComponentPackage, set *bundle.ComponentSe
 }
 
 // Validate validates components and components sets, providing as many errors as it can.
-func (v *ComponentValidator) Validate() []error {
-	var errs []error
+func (v *ComponentValidator) Validate() field.ErrorList {
```
mtaufen: What's the advantage of wrapping the things we want to validate in a `ComponentValidator`, vs. just having a validation function per type?

Kashomon: Good question. I think it's just ignorance on my part. What's the standard way to add these validation functions in K8S? Is it in the API directory? That's not ideal, because it'll lead to circular deps at the moment.

pkg/validation/validate_component.go
```diff
@@ -47,75 +50,108 @@ func NewComponentValidator(c []*bundle.ComponentPackage, set *bundle.ComponentSe
...
 
-func (v *ComponentValidator) validateComponentPackages() []error {
-	var errs []error
+func cPath(ref bundle.ComponentReference) *field.Path {
+	return field.NewPath(fmt.Sprintf("Component{%s, %s}", ref.ComponentName, ref.Version))
+}
+
+func (v *ComponentValidator) validateComponentPackages() field.ErrorList {
+	cpathIdx := func(idx int) *field.Path {
+		return field.NewPath("Component").Index(idx)
```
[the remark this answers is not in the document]
Kashomon: Correct, that's exactly what's going on. I thought of this but I wasn't sure how to do this better -- do you have any ideas?

pkg/validation/validate_component.go
```diff
@@ -47,75 +50,108 @@ func NewComponentValidator(c []*bundle.ComponentPackage, set *bundle.ComponentSe
...
 }
 
-func (v *ComponentValidator) validateComponentPackages() []error {
-	var errs []error
+func cPath(ref bundle.ComponentReference) *field.Path {
+	return field.NewPath(fmt.Sprintf("Component{%s, %s}", ref.ComponentName, ref.Version))
+}
+
+func (v *ComponentValidator) validateComponentPackages() field.ErrorList {
+	cpathIdx := func(idx int) *field.Path {
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/validation/validate_component.go
```diff
@@ -47,75 +50,108 @@ func NewComponentValidator(c []*bundle.ComponentPackage, set *bundle.ComponentSe
...
+
+	errs := field.ErrorList{}
 	objCollect := make(map[bundle.ComponentReference]bool)
-	for _, ca := range v.components {
+	for i, ca := range v.components {
 		n := ca.Spec.ComponentName
-		if err := validateName(n); err != nil {
-			errs = append(errs, fmt.Errorf("the component name %q was invalid config %v", n, ca))
+		if n == "" {
+			errs = append(errs, field.Required(cpathIdx(i).Child("Spec", "ComponentName"), "componentName is required"))
```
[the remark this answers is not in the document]
Kashomon: Done

pkg/validation/validate_component.go
Kashomon: uniqueness is a requirement

pkg/validation/validate_component.go
```diff
@@ -47,75 +50,108 @@ func NewComponentValidator(c []*bundle.ComponentPackage, set *bundle.ComponentSe
...
 			continue
 		}
 		objCollect[key] = true
 	}
 	return errs
 }
 
-func (v *ComponentValidator) validateComponentsSet() []error {
-	var errs []error
+func (v *ComponentValidator) validateComponentsSet() field.ErrorList {
```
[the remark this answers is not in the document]
Kashomon: Oops. Fixed.

pkg/validation/validate_component.go
mtaufen: Yeah basically `Validate{Type}(obj *Type) field.ErrorList` functions for top-level resource types (usually that's anything that has ObjectMeta), and you can also have `validate{EmbeddedType}(obj *EmbeddedType, p *field.Path) field.ErrorList` style helpers for substructures that let you thread the fieldpath through validation. 

If you always start from validating a root object where every field has a name (even implicit), you also don't have to worry about subscripting an anonymous field (from above comment), because you can just include the name in the path without ambiguity (comment above on minimum I think we could do if you can't do this).

pkg/validation/validate_component.go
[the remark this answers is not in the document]
mtaufen: I think I would just index an empty name, like `field.NewPath("").Index(idx)`, and make sure I wrap the ErrorList in a useful message when it's returned from `validateComponentPackages` (assuming you can't just validate root resources instead).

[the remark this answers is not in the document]
mtaufen: Which would at least make it clear that we don't know the name, rather than suggesting a name that may or may not be correct.

pkg/validation/validate_component.go
mtaufen: Were we testing for uniqueness? Maybe I missed that?

pkg/commands/validate/validate_test.go
```diff
@@ -16,52 +16,50 @@ package validate
...
-	errs []error
+	errs field.ErrorList
 }
 
-func (f *fakeBundleValidator) Validate() []error {
+func (f *fakeBundleValidator) Validate() field.ErrorList {
 	return f.errs
 }
 
 func TestRunValidate(t *testing.T) {
```
mtaufen: Another test you can do, which will quickly give you some more coverage, is to try to trigger every error in your validation, and make sure the total error count matches what you expect (it's kind of the opposite of a "set every field to something valid and ensure I get no errors" test).

pkg/validation/validate_component.go
mtaufen: Conventionally you would put the validation close to where the objects are defined. What are the circular deps at the moment?

pkg/validation/validate_component.go
```diff
@@ -47,75 +50,106 @@ func NewComponentValidator(c []*bundle.ComponentPackage, set *bundle.ComponentSe
...
+		p := cPath(ref)
+
+		if nameErrs := validateName(p.Child("Spec", "ComponentName"), n); len(errs) > 0 {
+			errs = append(errs, nameErrs...)
 		}
 
 		api := ca.APIVersion
 		if !apiVersionPattern.MatchString(api) {
-			errs = append(errs, fmt.Errorf("components APIVersion must have an apiVersion of the form \"bundle.gke.io/<version>\". was %q for config named %q", api, n))
+			errs = append(errs, field.Invalid(p.Child("APIVersion").Index(i), api, "must have the form \"bundle.gke.io/<version>\""))
```
mtaufen: Could use backticks not have to escape the quotes:
```
`must have the form "bundle.gke.io/<version>"`
```

pkg/validation/validate_component.go
```diff
@@ -126,44 +160,46 @@ func (v *ComponentValidator) validateComponentsSet() []error {
...
 			n := obj.GetName()
 			if n == "" {
-				errs = append(errs, fmt.Errorf("objects must always have a metadata.name. was empty object %d in component %q", i, compName))
+				errs = append(errs, field.Required(cPath(ref).Child("Spec", "Objects").Index(i).Child("Metadata", "Name"),
+					""))
 				continue
 			}
-			if err := validateName(n); err != nil {
-				errs = append(errs, fmt.Errorf("invalid name %q for object %d: %v", n, i, err))
+			p := cPath(ref).Child("Spec", fmt.Sprintf("Objects[%s]", n))
```
mtaufen: could use `.Key(string)`

pkg/validation/validate_component.go
```diff
@@ -126,44 +160,46 @@ func (v *ComponentValidator) validateComponentsSet() []error {
 	// It is possible for there to be components that the component set does not
 	// know about, but all components in the component set must be in the
 	// components list
-	for _, ref := range v.componentSet.Spec.Components {
+	for i, ref := range v.componentSet.Spec.Components {
 		if _, ok := compMap[ref]; !ok {
-			errs = append(errs, fmt.Errorf("could not find component reference %v for any of the components", ref))
+			errs = append(errs, field.Duplicate(p.Child("Spec", "Components").Index(i), fmt.Sprintf("component ref %v", ref)))
```
mtaufen: Should this be `Required` or `NotFound` instead of `Duplicate`?
Could also use `.Key(string)` here.

pkg/validation/validate_component.go
Kashomon: Good idea

pkg/validation/validate_component.go
Kashomon: Done

pkg/validation/validate_component.go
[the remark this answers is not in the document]
Kashomon: Good idea

pkg/validation/validate_component.go
Kashomon: Done

pkg/validation/validate_component.go
Kashomon: I kept around the validator object. I'm imagining that I'll want configurable validation, so it's nice to have this around. That being the case, I feel like having a single validator object is more flexible.

pkg/commands/validate/validate_test.go
Kashomon: Added

Kashomon:
> One thing for the future: If you're using K8s deserializers, you might not need to manually validate apiVersion and kind, since I think these are required for the deserializer to know the destination type.

ah, cool. thanks for the tip.

PR #118, opened 2018-12-19
Add an AppVersion field to the ComponentSpec.

Kashomon:
This PR adds AppVersion as a field to the component spec. Note that the
versioning is much less restrictive here than for the `version` field.
The app version will accept X.Y and X.Y.Z-blah.0 in addition to X.Y.Z

Note: This intentionally doesn't add requirements. I now see this more
as something that would live as a CRD in the component objects.

### DIFF
--- a/pkg/validate/validate_component.go
+++ b/pkg/validate/validate_component.go
@@ -27,14 +27,24 @@ import (
 var (
 	apiVersionPattern = regexp.MustCompile(`^bundle.gke.io/\w+$`)
 
-	// Regex string for numbers without leading zeros.
-	number = `([1-9]\d*|0)`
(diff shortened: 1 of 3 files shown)


pkg/validate/validate_component.go
```diff
@@ -28,13 +28,19 @@ var (
 	apiVersionPattern = regexp.MustCompile(`^bundle.gke.io/\w+$`)
 
 	// Regex string for numbers without leading zeros.
-	number = `([1-9]\d*|0)`
+	numPattern = `([1-9]\d*|0)`
+
+	extraVersionInfo = `-[a-zA-Z0-9_.+-]+`
```
BradfordMedeiros: If you could comment all the regex, that would help.

Issue #119, opened 2019-01-07
Add filename in converter.FromFileName


Kashomon:
Currently, `converter.FromFileName` has the following format:

```go
// FromFileName creates a Muxer instance by guessing the type based on the file
// extension.
func FromFileName(fname string, contents []byte) *Muxer {
	ext := filepath.Ext(fname)
	switch ext {
	case ".yaml":
		return FromYAML(contents)
	case ".json":
		return FromJSON(contents)
	default:
		// This will be an error during conversion
		return &Muxer{format: UnknownContent}
	}
}
```

In the case of an error, it should note the file name, since it has that information.

PR #122, opened 2019-01-10
Add filename when return error in converter.FromFileName #119

u5surf:
#119 
When return error in converter.FromFileName , it returns UnknownContent with filename

### DIFF
--- a/pkg/converter/muxer.go
+++ b/pkg/converter/muxer.go
@@ -63,7 +63,7 @@ func FromFileName(fname string, contents []byte) *Muxer {
 		return FromJSON(contents)
 	default:
 		// This will be an error during conversion
-		return &Muxer{format: UnknownContent}
+		return &Muxer{format: UnknownContent + ContentType("file: "+fname)}
 	}
 }
 


pkg/converter/muxer.go
```diff
@@ -63,7 +63,7 @@ func FromFileName(fname string, contents []byte) *Muxer {
 		return FromJSON(contents)
 	default:
 		// This will be an error during conversion
-		return &Muxer{format: UnknownContent}
+		return &Muxer{format: UnknownContent + ContentType("file: "+fname)}
```
Kashomon: It seems a little odd to reuse content type. Although it's true it'll get passed back as an error, I think it'll be confusing to the reader. I'd prefer that we add a new field that looks something like

```go
filesource string
```

If there's an error, we add that to the error message explicitly.

u5surf: I think it is just changed error messages following 

pkg/converter/muxer.go
```
default:
		// This will be an error during conversion
		return &Muxer{format: CotentType(fname +" has neither extend of .yaml nor .json")}
```
or, defined neatly variable `UnsupportedContentType` which called when errors

pkg/converter/converter.go
```
var (
  // UnknownContent indicates content of unknown type.
  UnknownContent ContentType

  // YAML indicates YAML content.
  YAML ContentType = "yaml"

  // JSON indicates JSON content.
  JSON ContentType = "json"
  // 
  UnsupportedContentType  ContentType = "Unspported file format"
)
```

pkg/converter/muxer.go
```
default:
		// This will be an error during conversion
		return &Muxer{format: UnsupportedContentType + ContentType(": " + fname) )}
```

Kashomon: Yeah, I realize that's technically what it's doing, but it doesn't seem quite right to overload this the UnsupportedContentType. It's basically an enum (and I would use enums if Go had them), so shouldn't change.

Kashomon:
This functionality was added via #171

PR #125, opened 2019-01-15
Add the BundleBuilder and rework some smaller API features

### LINKED ISSUE #120: API: Rename ComponentPackage to Component
Originally Component was called ClusterApplication. The it was called ClusterComponent, because the things are not necessarily applications. Then it was called ComponentPackage to make it clear that it's packaging scheme.

However, I've found that the Package bit introduces an extra noun, which causes confusing. If I say 'package' and 'component' do I just mean 'component'? 

So, the KISS approach says to me: Let's pick just one of Package or Component. I think Component is a little more clear than just Package.

Kashomon:
- ~~Add a BundleBuilder and ComponentBuilder types. BundleBuilders can
  reference files that are either components or component builders and moved all file-y concepts from Components/Bundles to the Builders.~~
- ~~Rename ComponentPackage to Component.~~
- ~~Added a bundle wrapper object for bundle input/output~~
- ~~Move package `inline` to `build` and reworked inlining to only work on the builder types. Additionally, rewrite the inlining tests.~~
- ~~Remove path rewriting -- make everything rely on absolute paths. This
  removes several hundred lines and I had already stopped using it.~~
- Fix examples to use the new builder pattern.

Fixes #120


ihmccreery:
This is a 4k+ line change; can it be broken up into smaller pieces?

Kashomon:
> This is a 4k+ line change; can it be broken up into smaller pieces?

There's a lot of generated code here, so it's bigger than it seems. Unfortunately, this felt like the smallest change, while still ensuring the repository still continued to build.

I could split out the ComponentPackage -> Component change if that would help.

ihmccreery:
> Unfortunately, this felt like the smallest change, while still ensuring the repository still continued to build.

That seems like a bad sign.  Is there some refactoring we can do before making this change, then?

> I could split out the ComponentPackage -> Component change if that would help.

That would help, I expect.

Kashomon:
I could also introduce the types (and not have them used) in a separate cl.

ihmccreery:
> I could also introduce the types (and not have them used) in a separate cl.

Sgtm.

In general, each of those bullets in the PR description look like they could be their own PR.

Kashomon:
* Moved adding bundle types in #129
* Moved adding bundle wrapper object to #130

Kashomon:
#132 - rework the inlining code 

This PR is almostly been completely submitted in children PRs. I don't think this PR will get merged.

Kashomon:
Abandoned -- moved to #135

PR #128, opened 2019-01-15
Rename ComponentPackage to Component

artasparks:
Having both Component and Package in the name has added significant
confusion. Is it a package? Component? Additionally, it was originally
added for compatibility with another project, but that other project is
no longer using the Package terminology.


johnbelamaric:
Fix the CI issue, otherwise looks good.

Kashomon:
Should be fixed with latest commim

Issue #116, opened 2018-12-13
API: Add Builder forms of ComponentPackage and ComponentSet/Bundle


Kashomon:
For the API, I'd love to add Builder forms of ComponentPackage and ComponentSet. The idea here is that these we move inlining out of the ComponentPackage and the Bundle. These become 'Builder' objects that get transformed into ComponentPackages and ComponentSets. This build phase is thus rather like a compile stage. It could even allow for customization (insertion of images).

Here's what I'm imagining

```go
type ComponentPackageBuilder struct {
        metav1.TypeMeta   `json:",inline"`
        metav1.ObjectMeta `json:"metadata,omitempty"`

        ComponentName string `json:"componentName,omitempty"`
        Version string `json:"version,omitempty"`

        ObjectFiles []File `json:"objectFiles,omitempty"`
        RawTextFiles []FileGroup `json:"rawTextFiles,omitempty"`
}

// === Outputs === >

type ComponentSet struct {
        metav1.TypeMeta   `json:",inline"`
        metav1.ObjectMeta `json:"metadata,omitempty"`
        Spec ComponentSetSpec `json:"spec,omitempty"`
}

type ComponentPackageSpec struct {
        ComponentName string `json:"componentName,omitempty"`
        Version string `json:"version,omitempty"`
        Objects []*unstructured.Unstructured `json:"objects,omitempty"`
}
```

And, for sets and bundles.

```go
type BundleBuilder struct {
        SetName string `json:"setName,omitempty"`
        Version string `json:"version,omitempty"`
        ComponentBuilderFiles []File `json:"componentFiles,omitempty"`
}

// === Outputs === >

type ComponentSetSpec struct {
        SetName string `json:"setName,omitempty"`
        Version string `json:"version,omitempty"`
        Components []ComponentReference `json:"components,omitempty"`
}


type ComponentSet struct {
        metav1.TypeMeta   `json:",inline"`
        metav1.ObjectMeta `json:"metadata,omitempty"`
        Spec ComponentSetSpec `json:"spec,omitempty"`
}

// And, a full, thin Bundle

type Bundle struct {
        metav1.TypeMeta   `json:",inline"`
        metav1.ObjectMeta `json:"metadata,omitempty"`
        SetName string `json:"setName,omitempty"`
        Version string `json:"version,omitempty"`
        Components []*ComponentPackage `json:"components,omitempty"`
}
```

PR #129, opened 2019-01-16
Add BundleBuilder and ComponentBuilder Types.

Kashomon:
The builder types contain just the file-references from the
Component/Bundles. When finished, the bundle and component objects will
just have the inlined forms.

Incremental progress on #116


Kashomon:
Once this is in it should make it easier to split up some more of the other things too

PR #130, opened 2019-01-16
Add bundlewrapper object for bundle I/O

Kashomon:
The BundleWrapper object is a simple union type that's useful for doing
input/output of bundle objects. It contains some logic for ensuring that
only one of the various bundle objects is non-nil at a given time.

- Incremental progress on #116
- Depends on #129


Kashomon:
This will fail until rebased on #129

Issue #98, opened 2018-11-25
Testing: Round-Trip Converter Tests


Kashomon:
We should add tests to converter library to ensure that round-trip conversion works and gives the same result (->ToObject->ToYAML->ToObject). It would help catch issues like https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/pull/93

Zyqsempai:
I will take care of it!

PR #131, opened 2019-01-16
Added one more convert iteration from Yaml to Object

Zyqsempai:
Added one more convert iteration from Yaml to Object.
Added few checks to be sure that Object field value matches original YAML values.

Fixes https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/98

### DIFF
--- a/pkg/converter/converter_test.go
+++ b/pkg/converter/converter_test.go
@@ -57,13 +57,17 @@ spec:
 `
 
 func TestConvertUnstructured(t *testing.T) {
-	p, err := FromYAMLString(reschedulerManifest).ToUnstructured()
+	firstIttObj, err := FromYAMLString(reschedulerManifest).ToUnstructured()
 	if err != nil {
 		t.Fatalf("unexpected error parsing manifest: %v", err)
 	}
-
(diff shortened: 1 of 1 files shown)


Zyqsempai:
cc @Kashomon

pkg/converter/converter_test.go
```diff
@@ -57,13 +57,17 @@ spec:
 `
 
 func TestConvertUnstructured(t *testing.T) {
```
Kashomon: I think it would be useful to also try to do this for the bundle / component objects. Maybe better as a follow-up PR

Kashomon:
Oops, wrong command.
/ok-to-test

Kashomon:
Thanks for the PR!

pkg/converter/converter_test.go
borispopovschi-newstore: You can open an issue, and assign me on it. Will be glad to continue working.

Zyqsempai: You can open issue and assign me on it, will be glad to continue working.

PR #132, opened 2019-01-17
Copy the inline code to pkg/build and rework to use Builder forms.

Kashomon:
This PR reworks the inlining code in the following ways:

- Move to pkg/build, to set up for more building functionality.
- Remove relative path rewriting. The files must now all have absolute paths.
- Use the BundleBuilder/ComponentBuilder types, which get inlined to Bundle, Component types.
- Rewrite the tests from scratch to use a table-driven approach.

This was split out from #125
And is incremental progress on #116

Note: This has the downside of duplicating code between here and the
`pkg/inline` directory and also making the diffs / blame less clear. But
I think it's a bit easier to do incrementally this way.

### DIFF
--- a/pkg/build/config_map_maker.go
+++ b/pkg/build/config_map_maker.go
@@ -1,0 +1,67 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/converter"
+)
+
+type configMapMaker struct {
+	cfgMap *corev1.ConfigMap
+}
+
+// Make a new ConfigMap with a metdata.name.
+//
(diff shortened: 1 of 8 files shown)


pkg/build/config_map_maker.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2018 Google LLC
```
Kashomon: This file did not change

pkg/build/config_map_maker_test.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2018 Google LLC
```
Kashomon: this file did not change

pkg/build/inline.go
```diff
@@ -0,0 +1,223 @@
...
+	rdrMap := map[files.URLScheme]files.FileObjReader{
+		scheme: objReader,
+	}
+	return &Inliner{
+		Readers: rdrMap,
+	}
+}
+
+// BundleFiles converts dereferences file-references in for bundle files.
+func (n *Inliner) BundleFiles(ctx context.Context, data *bundle.BundleBuilder) (*bundle.Bundle, error) {
```
Kashomon: BundleFiles was rewritten

pkg/build/inline.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	return newBundle, nil
+}
+
+var onlyWhitespace = regexp.MustCompile(`^\s*$`)
+var multiDoc = regexp.MustCompile("---(\n|$)")
+
+// ComponentFiles reads file-references for component builder objects.
+// The returned components are copies with the file-references removed.
+func (n *Inliner) ComponentFiles(ctx context.Context, comp *bundle.ComponentBuilder) (*bundle.Component, error) {
```
Kashomon: Not much changed in ComponentFiles

pkg/build/inline.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, newc)
+	}
+	return out, nil
+}
+
+// readFile from either a local or remote location.
+func (n *Inliner) readFile(ctx context.Context, file bundle.File) ([]byte, error) {
```
Kashomon: This did not change

Kashomon:
I've tried to indicate where there are changes

Kashomon:
I could leave the map-maker code in `pkg/inline` to make it clear it hasn't changed, if that's easier.

Issue #109, opened 2018-11-29
Reintroduce Patch Object, Maker


Kashomon:
I've added and removed the Patch objects a couple times. Now that we have the `pkg/maker` concept/directory, I believe it's time to reintegrate this work.

Kashomon:
I currently have a pending PR that introduces this to pkg/options/patch

PR #136, opened 2019-01-24
Add a new patch options applier.

Kashomon:
This PR reworks patching and adds a bunch of tests.

- Add a new minimal PatchTemplate type.
- Add a patch applier package
- Move the scheme logic to the patch applier

The patch logic has some quirks:

- All templates that are part of the the PatchTemplate must include a
  Kind, and apiversion (I think if it's not a v1 type).
- Additionally, suppling a name for the Template will match against the
  name of potential match-objects
- The applier can take a 'filter' object to filter the PatchTemplate
  objects. This is useful for if two patches apply to a single object, but
  for example, you want one of the two patches to apply.

Fixes: #109


johnbelamaric:
This looks great, it will be super-useful. How do the matching and syntax (e.g., $patch: delete) compare to `kustomize`? Are they the same or similar?

Kashomon:
`$delete` is strategic merge patch logic. I know that` kustomize` uses the same strategic merge patch library, but I don't know if they support this directive (or prohibit it, rather). I've talked to some API Machinery folks who are pretty against delete because it's not always possible via `kubectl`, since some fields are readonly. 

Personally, I think it's useful from a build perspective, but maybe shouldn't be supported from a runtime perspective.

johnbelamaric:
Ok, cool.

/lgtm

PR #145, opened 2019-01-30
Add a patch template builder type.

Kashomon:
This adds a patch builder type. The big change is this adds two schemas
to the builder (before/after) and adds one schema to the resulting
PatchTemplate.

The idea is that at the same time that inlining happens (or slightly
after), PatchTemplateBuilders could be converted to PatchTemplates.

Here's a worked example:

```yaml

  apiVersion: bundleext.gke.io/v1alpha1
  kind: PatchTemplateBuilder
  template: |
    apiVersion: v1
    kind: Pod
    spec:
      containers:
        - name: foo-container
          image: '{{.Registry}}/{{.Repository}}/{{.Image}}:{{.Tag}}'

  buildSchema
    # Properties that are not filled out during build-time
    Registry:
      default: '{{.Registry}}'
    Repository:
      default: '{{.Repository}}'

    # Properties that must be supplied during build
    Image:
      type: string
    Tag:
      type: string

  targetSchema:
    # Properties that must be filled out during runtime
    Registry:
      type: string
    Repository:
      type: string
```

Another possibility would be to treat PatchTemplates themselves as
'partials'.  I might prefer this option, but I haven't totally thought
it through. Maybe you could have a PartialApply method or command.


Kashomon:
After thinking about it some more, I think it's now possible to have a partial-options applier for patches.

Kashomon:
Nope, the partial patch attempt failed. I think you do need a builder.

Kashomon:
Reverting this since I'm merging in pieces.

PR #148, opened 2019-02-01
Add a patch template builder type + options schema

Kashomon:
This PR adds

- A builder form of the PatchTemplate
- A schema for the Patch Template options
- Two schemas for the PatchTemplateBuilder (before/after)

Note: The schema is the same used by CRDs, so we will reuse the same logic.

Partial work on #145


johnbelamaric:
/lgtm

Ok, I think this is all coming together nicely.

Issue #141, opened 2019-01-29
Add a PatchTemplateBuilder


Kashomon:
As we have a ComponentBuilder and BundleBuilder, it would be useful to have PatchTemplateBuilder. This would create partial or fully realized patches as part of the build process.

Here's a worked example of what I'm talking about:

```
  apiVersion: bundleext.gke.io/v1alpha1
  kind: PatchTemplateBuilder
  template: |
    apiVersion: v1
    kind: Pod
    spec:
      containers:
        - name: foo-container
          image: '{{.Registry}}/{{.Repository}}/{{.Image}}:{{.Tag}}'

  buildSchema
    # Properties that are not filled out during build-time
    Registry:
      default: '{{.Registry}}'
    Repository:
      default: '{{.Repository}}'

    # Properties that must be supplied during build
    Image:
      type: string
    Tag:
      type: string

  targetSchema:
    # Properties that must be filled out during runtime
    Registry:
      type: string
    Repository:
      type: string
```

Kashomon:
This is finished now.

PR #152, opened 2019-02-06
Add openapi validation method.

### LINKED ISSUE #2: Add validation for CRDs
We should add validation for CRDs. The Kubernetes libraries provide this functionality, but I held off on this due to import-dependency-issues.

Kashomon:
This method uses OpenAPI schemas to validate arbitrary JSON structures.
I moved this to an openapi directory from validate because this is also
where I will put defaulting logic. The openapi library takes the result
of a validate call and allows you to use it to other transformations
(e.g., defaulting).

Addresses #141 
Fixes #2


pkg/options/openapi/validate.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2018 Google LLC
```
johnbelamaric: 2019

pkg/options/openapi/validate_test.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2018 Google LLC
```
johnbelamaric: 2019

pkg/options/openapi/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
johnbelamaric: Why do we need this here? We already have it at the root level.

Kashomon: Well, I copied this file wholesale from K8S, so I felt like I needed to copy the license. WDYT?

johnbelamaric:
No, I don't think so. If we had a different license, then possibly. But we
already have the same license. You have preserved the copyright notices,
that is sufficient.

pkg/options/openapi/LICENSE
Kashomon: Removed.

pkg/options/openapi/validate.go
Kashomon: Done

pkg/options/openapi/validate_test.go
Kashomon: done

Kashomon:
PTAL, fixed the copyright/license issues

PR #162, opened 2019-02-07
Add support for PatchTemplateBuilder

johnbelamaric:
This adds building PatchTemplateBuilder resources into PatchTemplate resources to `bundlectl build`.

It also enables schema validation and defaulting in `bundlectl patch`.


pkg/commands/build/build.go
```diff
@@ -21,23 +21,34 @@ import (
...
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/filter"
 	log "github.com/golang/glog"
 	"github.com/spf13/cobra"
 )
 
 // options represents options flags for the build command.
-type options struct{}
+type options struct {
+	// annotations selects a subset of patch templates on which to apply options
+	// Has the form "foo,bar;biff,baz".
```
Kashomon: I might go through and change all these to foo=bar,biff=baz. But I'll do that in a following PR.

pkg/commands/build/build.go
```diff
@@ -48,11 +59,29 @@ var createInlinerFn = func(pbr files.FileObjReader) *build.Inliner {
...
+func run(ctx context.Context, o *options, brw cmdlib.BundleReaderWriter, rw files.FileReaderWriter, gopt *cmdlib.GlobalOptions) error {
 	bw, err := brw.ReadBundleData(ctx, gopt)
 	if err != nil {
 		return fmt.Errorf("error reading bundle contents: %v", err)
 	}
 
+	//TODO(jbelamaric): allow multiple files
+	// the bundle now contains components which may include PatchTemplateBuilder objects
+	// that we need to build into PatchTemplates
+        optFiles := []string{}
```
Kashomon: formatting's a bit weird here

pkg/commands/cmdlib/bundleio.go
```diff
@@ -202,3 +204,40 @@ func formatFromFile(path string) string {
...
+		kv := strings.Split(v, ",")
+		if len(kv) == 2 {
+			m[kv[0]] = kv[1]
+		}
+	}
+	return m
+}
+
+// MergeOptions reads multiple options files and combines them into a single map.
+// Option values will be overwritten if a later file has the same key.
```
Kashomon: I think it's probably clearer to produce an error if there's a duplicate key. imo, it's confusing behavior to overwrite.

pkg/filter/filter.go
```diff
@@ -99,11 +99,22 @@ func (f *Filter) Components(data []*bundle.Component, o *Options) []*bundle.Comp
...
+	matched, notMatched := f.PartitionObjects(data, o)
+
+	if o == nil || o.KeepOnly {
+		return matched
+	}
+
+	return notMatched
+}
+
+// PartitionObjects splits the objects into matched and not matched sets.
```
Kashomon: Nice addition.

examples/patchbuilder/patch-template-builder.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - K8sVersion
+  properties:
+    k8sVersion:
+      description: The kubernetes version for this build. This will be used as the image tag for Kubernetes images.
+      type: string
+targetSchema:
+  properties:
+    Registry:
+      description: The registry from which the image should be pulled. Should include
+        the trailing /.
```
Kashomon: why include the trailing /?

pkg/build/patchbuild.go
```diff
@@ -0,0 +1,147 @@
...
+	if ptb.BuildSchema != nil {
+		opts, err = openapi.ApplyDefaults(opts, ptb.BuildSchema)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// This is a hack to allow us to pass through runtime templates variables
+	if ptb.TargetSchema != nil && ptb.TargetSchema.Properties != nil {
+		for k := range ptb.TargetSchema.Properties {
```
Kashomon: This only looks one-layer down, but you could have nested properties structure. We should support nested structures. You can file a TODO to update this behavior and we can do it in a follow-up PR.

pkg/build/patchbuild.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+		ObjectMeta:    *ptb.ObjectMeta.DeepCopy(),
+		OptionsSchema: ptb.TargetSchema.DeepCopy(),
+		Template:      buf.String(),
+	}
+	return pt, nil
+}
+
+func BuildComponentPatchTemplates(c *bundle.Component, fopts *filter.Options, opts options.JSONOptions) (*bundle.Component, error) {
+	log.Info("Building PatchTemplates for component")
```
Kashomon: nit: I don't think this log line adds too much as is. If you want to keep it, probably needs the component name / version.

pkg/build/patchbuild.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			return nil, err
+		}
+		c.Spec.Objects = append(c.Spec.Objects, o)
+	}
+
+	return c, nil
+}
+
+func BuildAllPatchTemplates(bw *wrapper.BundleWrapper, fopts *filter.Options, opts options.JSONOptions) (*wrapper.BundleWrapper, error) {
```
Kashomon: needs comment

pkg/build/patchbuild.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/converter"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/filter"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/options"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/options/openapi"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/wrapper"
+	log "github.com/golang/glog"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ComponentFiles reads file-references for component builder objects.
```
Kashomon: fix comment

pkg/build/patchbuild.go
```diff
@@ -0,0 +1,147 @@
...
+			Kind:       "PatchTemplate",
+		},
+		ObjectMeta:    *ptb.ObjectMeta.DeepCopy(),
+		OptionsSchema: ptb.TargetSchema.DeepCopy(),
+		Template:      buf.String(),
+	}
+	return pt, nil
+}
+
+func BuildComponentPatchTemplates(c *bundle.Component, fopts *filter.Options, opts options.JSONOptions) (*bundle.Component, error) {
```
Kashomon: needs comment

pkg/build/patchbuild.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if ptb.BuildSchema != nil {
+		opts, err = openapi.ApplyDefaults(opts, ptb.BuildSchema)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// This is a hack to allow us to pass through runtime templates variables
```
Kashomon: I haven't thought of a good non-hacky way to do this. I'm all ears =)

pkg/commands/cmdlib/bundleio.go
```diff
@@ -202,3 +204,40 @@ func formatFromFile(path string) string {
 	}
 	return ""
 }
+
+// ParseAnnotations parses CLI flag format for annotations
+func ParseAnnotations(p string) map[string]string {
```
Kashomon: nit: really something like ParseKeyValueFlag

pkg/filter/filter.go
```diff
@@ -99,11 +99,22 @@ func (f *Filter) Components(data []*bundle.Component, o *Options) []*bundle.Comp
...
+func (f *Filter) PartitionObjects(data []*unstructured.Unstructured, o *Options) ([]*unstructured.Unstructured, []*unstructured.Unstructured) {
```
Kashomon: Needs a test. Can you add a TODO for me?

examples/patchbuilder/patch-template-builder.yaml
johnbelamaric: So you can leave it bare if you want. But since I ended up putting the default in here, I don't think you can do that anyway. So I will remove it.

pkg/build/patchbuild.go
johnbelamaric: done

pkg/build/patchbuild.go
johnbelamaric: Me either! But I realize that this ONLY works for simple substitutions. I can't do, for example

```
{{- if .ImagePullSecret != "" }}imagePullSecrets:
- name: {{.ImagePullSecret}}{{end}}
```

Because the "build" time template execution will execute the `if`.

pkg/build/patchbuild.go
johnbelamaric: Ahh. OK.

pkg/build/patchbuild.go
johnbelamaric: gone

pkg/build/patchbuild.go
johnbelamaric: done

pkg/build/patchbuild.go
johnbelamaric: done

pkg/commands/build/build.go
Kashomon: Filed as #163

pkg/commands/build/build.go
johnbelamaric: i just removed the TODO, it's already captured elsewhere

pkg/commands/cmdlib/bundleio.go
johnbelamaric: Well, I am thinking for this for specialization. A global options file, then individual options files that may override the global ones.

pkg/filter/filter.go
johnbelamaric: for you...TODO(jhoak)...or did you mean for me :) ?

pkg/commands/cmdlib/bundleio.go
johnbelamaric: I left it for now, but maybe if you change to the "=" format. By the way, in patch applier you have a comment about whether a whole filter is needed or just annotations. I think just annotations may be cleaner. What happens if they have a filter with "kind"? it have unexpected effects.

I followed the example and did a full filter, but we may want to change that.

pkg/filter/filter.go
Kashomon: me. not a big deal either way.

Kashomon:
Make sure to rebase before you commit (or squash and merge)

PR #164, opened 2019-02-08
Implement stripping applied patches from the component

johnbelamaric:
With this change, any patch that has been applied to the component will be removed from it by default. It is override-able with a flag.


pkg/options/patchtmpl/patch.go
```diff
@@ -75,24 +82,28 @@ func (p *parsedPatch) String() string {
 	return string(p.raw)
 }
 
-func (a *applier) makePatches(comp *bundle.Component, opts options.JSONOptions) ([]*parsedPatch, error) {
+func (a *applier) makePatches(comp *bundle.Component, opts options.JSONOptions) ([]*parsedPatch, []*unstructured.Unstructured, error) {
```
Kashomon: nit: would be helpful to add a comment about this does, especially what the return types are (what are these unstructured objects?)

PR #165, opened 2019-02-08
Implement the export command.

johnbelamaric:
This command will reads in a bundle or component and outputs a
--- delimted YAML file with all of the objects.


pkg/commands/export/export.go
```diff
@@ -0,0 +1,69 @@
...
+// options represents options flags for the export command.
+type options struct{}
+
+// opts is a global options instance for reference via the add commands.
+var opts = &options{}
+
+func action(ctx context.Context, cmd *cobra.Command, _ []string) {
+	gopt := cmdlib.GlobalOptionsValues.Copy()
+	stdio := cmdlib.RealStdioReaderWriter{}
+	brw := cmdlib.NewBundleReaderWriter(
```
BradfordMedeiros: No actions needed, but seeing this form being written every new command is a smell.  I'll take care of cleaning this up.

pkg/commands/export/export.go
```diff
@@ -0,0 +1,69 @@
...
+	log "github.com/golang/glog"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// options represents options flags for the export command.
+type options struct{}
+
+// opts is a global options instance for reference via the add commands.
+var opts = &options{}
```
BradfordMedeiros: again, it's the same form as other code, but this static options pattern is probably non ideal.  We could just be passing the options directly instead of relying on this weird static functionality.

pkg/commands/export/export.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func run(ctx context.Context, o *options, brw cmdlib.BundleReaderWriter, stdio cmdlib.RealStdioReaderWriter, gopt *cmdlib.GlobalOptions) error {
+	bw, err := brw.ReadBundleData(ctx, gopt)
+	if err != nil {
+		return fmt.Errorf("error reading bundle contents: %v", err)
+	}
+
+	comps := bw.AllComponents()
+	if len(comps) == 0 {
```
BradfordMedeiros: Why is len(comps) == 0 a special case?  To me working on a length 0 set still makes sense.

pkg/commands/export/export.go
```diff
@@ -0,0 +1,69 @@
...
+	brw := cmdlib.NewBundleReaderWriter(
+		&files.LocalFileSystemReaderWriter{},
+		&stdio,
+	)
+	if err := run(ctx, opts, brw, stdio, gopt); err != nil {
+		log.Exit(err)
+	}
+}
+
+func run(ctx context.Context, o *options, brw cmdlib.BundleReaderWriter, stdio cmdlib.RealStdioReaderWriter, gopt *cmdlib.GlobalOptions) error {
```
BradfordMedeiros: The only thing I'd like to see added before this code is merged, and I would highly appreciate, is just some example input files and the expected output.   This could be files of the expected input vs output, a formal test, just some comments inline here that show the form -- whatever works for you.

johnbelamaric: Take a look at the example I just added. It doesn't show the output directly but does show the commands.

pkg/commands/export/export.go
johnbelamaric: well, it still "works" as is, it just shortcuts in this case since there won't be any actual output. I thought ObjectExporter might puke on nil objects, so it was either this or put a check in there, or intialize objs to an empty slice.

PR #167, opened 2019-02-09
Url validation

BradfordMedeiros:
Got rid of dead url validation code, inliner fn dead code.


pkg/validate/validate_url.go
```diff
@@ -1,34 +0,0 @@
-// Copyright 2018 Google LLC
```
Kashomon: Can you revert the validate_url deletion? I'd like this to be used to validate component builder[s]

BradfordMedeiros: The abstraction was unclear to me so let's talk about where this belongs:

https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/commands/cmdlib/bundleio.go#L143

Bundle io calls the inliners to read from the file system to read files, which in turn call readFile which call the mappers for the scheme, see:
https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/build/inline.go#L222

Where do you see validation sitting?  To me validation would need one per scheme, correct?   Is the desire to add validation prior toward attempting toward reading the inlined file?

BradfordMedeiros: and if the validation is the same for each scheme type, i would assume: 
https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/build/inline.go#L223

Is the correct validation function, and we can use it for each file inliner.

Is my thinking correct, or am I missing something?

Kashomon: We would need one validation func per type, yes.

> Where do you see validation sitting?

The idea is that you should be able to validate structural properties of your file before inlining. This allows us to write simple presubmits that ensure the file is set up correctly.

In the language of Skylark (Bazel), you might imagine using such a validate command to make a rule.

`
validate_component(
  name = "my-component"
  srcs = ["foo.yaml"]
)

BradfordMedeiros: Makes sense for the intended usage but:
I see validation has a bunch of unused functions by the rest of the code. 

I assume this is being used elsewhere as a library, but then this is a distinct codepath from the actual reading and manipulation, which means we're maintaining multiple codepaths in which the validation can be different than the actual implementation of reading the files. Is that correct?  

For example, if I change validateName and disallow a certain type of string, there's no actual logic to integrate that with the default parsing that you hit in the build codepath?

Apologies if I'm overlooking something obvious.

Kashomon: Currently, it's just used as part of the `validate` command, it's true.

However, we should probably use it as the first step for all commands.

PR #168, opened 2019-02-11
Add support for nested properties during patch template building.

Kashomon:
This PR recursively looks through the target schema and creates template-parameters based on the keys that it finds in the `properties`.


johnbelamaric:
Nice

/lgtm

Kashomon:
Update was just comments changes. Merging.

PR #171, opened 2019-02-12
Encoder/decoder cleanups.

Kashomon:
Fixes: #119

* Adds filename to error messages during encodeng.
* Removes demux/mux private methods that I missed
* Add slightly more helpful error messages in converting library
* Update inlining error messages during conversion.
* Clean up existing conversion error message usage.

### DIFF
--- a/pkg/converter/decoder.go
+++ b/pkg/converter/decoder.go
@@ -57,15 +57,18 @@ func FromJSON(b []byte) *Decoder {
 // extension.
 func FromFileName(fname string, contents []byte) *Decoder {
 	ext := filepath.Ext(fname)
+	var d *Decoder
 	switch ext {
 	case ".yaml", ".yml":
-		return FromYAML(contents)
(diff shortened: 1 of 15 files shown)


pkg/converter/decoder.go
```diff
@@ -57,15 +57,18 @@ func FromJSON(b []byte) *Decoder {
 // extension.
 func FromFileName(fname string, contents []byte) *Decoder {
 	ext := filepath.Ext(fname)
+	var d *Decoder
```
BradfordMedeiros: I know you're a heavy vimmer, how do you deal with these short variable names I'm seeing used so often around here?   
breaks all the ways I know how to use my tools

Kashomon: ha. I don't use refactoring tools very often. if I'm using a single-variable name, it's because it's very locally scoped. Which tools does it break, out of curiosity?

PR #172, opened 2019-02-12
Implement export of Bundle to ComponentSet

johnbelamaric:
* Change default component name to include component set name and
version when in a component set (prevents component sets from
overwriting other component sets components)

* Move Export code into wrapper so that it can be used when using this
code as a library rather than using the CLI.

* Exporting a Bundle now results in a ComponentSet and Components, rather than all objects of all components in the bundle.

### DIFF
--- a/pkg/build/inline.go
+++ b/pkg/build/inline.go
@@ -84,7 +84,7 @@ func (n *Inliner) BundleFiles(ctx context.Context, data *bundle.BundleBuilder) (
 			if err != nil {
 				return nil, fmt.Errorf("error converting file %q to a component builder: %v", f.URL, err)
 			}
-			if c.GetName() == "" {
+			if c.GetName() == "" && data.ComponentNamePolicy == "SetAndComponent" {
 				c.ObjectMeta.Name = strings.Join([]string{data.SetName, data.Version, c.ComponentName, c.Version}, "-")
 			}
 			compbs = append(compbs, c)
--- a/pkg/build/inline.go
+++ b/pkg/build/inline.go
@@ -84,6 +84,9 @@ func (n *Inliner) BundleFiles(ctx context.Context, data *bundle.BundleBuilder) (
 			if err != nil {
 				return nil, fmt.Errorf("error converting file %q to a component builder: %v", f.URL, err)
 			}
+			if c.GetName() == "" {
+				c.ObjectMeta.Name = strings.Join([]string{data.SetName, data.Version, c.ComponentName, c.Version}, "-")
+			}
 			compbs = append(compbs, c)
 		default:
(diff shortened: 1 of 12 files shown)


pkg/build/inline.go
```diff
@@ -84,6 +84,9 @@ func (n *Inliner) BundleFiles(ctx context.Context, data *bundle.BundleBuilder) (
 			if err != nil {
 				return nil, fmt.Errorf("error converting file %q to a component builder: %v", f.URL, err)
 			}
+			if c.GetName() == "" {
+				c.ObjectMeta.Name = strings.Join([]string{data.SetName, data.Version, c.ComponentName, c.Version}, "-")
```
Kashomon: is this naming the component with the set name?

johnbelamaric: Yes. When it is exported as part of a set. The reason is that if we have two different sets that create components with the same name, they can conflict with one another - one may overwrite the other.

johnbelamaric: Alternately, this can be a validation of some sort - that a component can't be replaced by different bundle/component set. I am trying to prevent screw ups - for example, two bundles that have an "nginx" component. If we don't metadata.name components uniquely across component sets, we're risky some nasty issues.

Maybe in the component set/bundle  case we *shouldn't* use a reference but instead embed the components. Thoughts?

Kashomon: > The reason is that if we have two different sets that create components with the same name, they can conflict with one another - one may overwrite the other.

My goal is that we develop automation and process so that we ensure that components are always named+versioned uniquely. So, it should be a no-op to replace a component with another component of the same name and version. Otherwise, we're breaking the contract of having component+version being a unique identifier for a component, which is one of the visions of the schema.

I understand your concern about safety and I think it might be fair to provide some duct tape while we don't have the automation+process. Is there a way to avoid forcing this as the default behavior? Maybe as an option?

pkg/wrapper/bundlewrapper.go
```diff
@@ -148,3 +149,36 @@ func (bw *BundleWrapper) AllComponents() []*bundle.Component {
 	}
 	return nil
 }
+
+func (bw *BundleWrapper) ExportAsObjects() ([]*unstructured.Unstructured, error) {
```
Kashomon: + comment

pkg/build/inline.go
johnbelamaric: I can make it an option in the BundleBuilder.

pkg/wrapper/bundlewrapper.go
johnbelamaric: done

pkg/build/inline.go
johnbelamaric: done

PR #174, opened 2019-02-13
Add validation url

BradfordMedeiros:
Just a minor PR. 

So this code is unused but the intent is to add it into the bundle cli validate path properly. 

- All is an arbitrary grouping.  Deleted. 
- Adding componentObjects validation to AllComponents, because it was missing.  Since this is a subpath of the type, it should be validated
- Got rid of ComponentsAndComponentSet.  We'll probably need this check later, and we can revive it from git then, but as of now, it lacks context.


BradfordMedeiros:
As a general comment, I'm seeing bits and pieces of unused code here and there.  
My coding style generally is nuke the unused bits first, and then as we need it restore it from source control.

I find it hard to reason about how to best maintain code that's not actually being hit.  

Feel free to reject the PR if we really think we want to maintain and then integrate these pieces as it comes up, maybe it's just a different top/down/button up

pkg/validate/validate_component.go
```diff
@@ -227,3 +185,7 @@ func ComponentObjects(cp *bundle.Component) field.ErrorList {
 	}
 	return errs
 }
+
+func cPath(ref bundle.ComponentReference) *field.Path {
+	return field.NewPath("Component").Key(fmt.Sprintf("%v", ref))
+}
```
Kashomon: nit: seems to be missing new line at the end

Kashomon:
looks good, but some tests are failing.

BradfordMedeiros:
Going slow, I pushed but don't mean to review it yet.  
I'm basically just separating the validate out to explicit ValidateComponent, ValidateComponentSet, and ValidateComponentSetForComponent (since the relation is relevant).

Found some interesting cases where validation seemed to be slightly off

pkg/validate/validate_component_test.go
```diff
@@ -173,21 +92,33 @@ components:
...
+func assertValidateComponent(componentString string, numErrors int, t *testing.T){
+	component, _ := converter.FromYAMLString(componentString).ToComponent()
+	errs := Component(component)
+	numErrorsActual := len(errs)
+	if numErrorsActual != numErrors {
+		t.Fatalf("expected %v errors got %v", numErrors, numErrorsActual)
+	}
+}	
+
+func TestComponentVersionDotNotationAppVersion(t *testing.T){
```
Kashomon: why split up the table test?

pkg/validate/validate_component_test.go
```diff
@@ -1,46 +1,10 @@
-// Copyright 2018 Google LLC
```
Kashomon: you lost the copyright

pkg/validate/validate_component_test.go
```diff
@@ -60,107 +24,62 @@ components:
...
-    version: 2.10.1
-    appVersion: 3.10.1`,
-		},
-		{
+    component, errConv := converter.FromYAMLString(defaultComponentNoRefs).ToComponent()
+    if errConv != nil {
+    	t.Fatalf("error converting ", errConv)
+    	return
+    }
+    errs := Component(component)
```
Kashomon: Your formatting is funny. Can you make sure you have gofmt auto-enabled on write or similar?

pkg/validate/validate_component_test.go
Kashomon: Personally I would prefer to have the refactoring the tests in a separate PR regardless. Table tests are pretty standard go-style, so I'm not sure what we gain from this  (TestComponentVersionDotNotationAppVersion feels very Java-y).

Kashomon: https://github.com/golang/go/wiki/TableDrivenTests

pkg/validate/validate_component_test.go
BradfordMedeiros: Still working on this.  Just pushed and it got updated.  Thought it would only update my branch and not show here.  I'll close the PR and reopen when I'm ready for review. 

I didn't intend to merge this as is.

pkg/validate/validate_component_test.go
BradfordMedeiros: So again, didn't mean to push this yet, but the old method All function was convoluting ideas as far as I could tell.  I was separating them out and pulling them out into individual tests was just easier.

Kashomon:
I saw a couple more commits come through -- Let me know when you think this is ready to review.

I'm still fairly opposed to replacing the table-tests with a single-test-per case WithAVerylongNameThatDescribesTheTest, since I don't feel like that's go-style. I think a good compromise could be having the structure still be a table test but then using your helper function in the table test. Let's chat if you feel strongly about that.

BradfordMedeiros:
I will revert back to the table test.  It's not really any functional difference so I don't care.   It just was separated because it was easier for me to pull out the tests this way.

BradfordMedeiros:
@Kashomon  This should be good to go now.
Took longer than I wanted (i was just slow no good reason -- you know those days :cry:  ), but I was just trying to separate the paths between validate component, validate component set, etc.

From my perspective the validate code path ought to be   something like in pseudo
~~~~
validate(bundleIo.readBundle(file).ToComponent()).
~~~~
which is just to say that for the corresponding bundle wrapper method there should be a corresponding validate.  Prior it was sort of a mixed thing which was a bit hard to reason about.

The structure would to be simply call the corresponding method (or we have a wrapper around the type) that calls the appropriate validate. 

It's worth noting that the original branch name is validate url.  Since we do not have a validate component builder function and components do not have urls, the validate url continues to be unused.

Kashomon:
Looks like the tests tests are still failing?

BradfordMedeiros:
Tests are failing in an interesting place.  I assume the old validation code is broken. 

It's failing because now we validate the component objects in the component. 
The objects validation validates the metadata.name exists.   This part is unstructured and is supposed to be able to represent any resource type. 
 
Is there a reason we verify the name field in the first place?   Should it be made required, and taken out? 

For example it fails here (this an element of the objects array of a component)
~~~~
 - apiVersion: bundle.gke.io/v1alpha1
    kind: PatchTemplate
    metadata:
      annotations:
        bundle.gke.io/inline-type: kube-object
    template: |
      apiVersion: v1
      kind: Pod
      metadata:
        namespace:
          {{.namespace}}
~~~~

Kashomon:
I think it's fine to remove the name validation.

BradfordMedeiros:
@Kashomon anything else on this PR?

pkg/commands/validate/validate.go
```diff
@@ -45,20 +43,23 @@ func action(ctx context.Context, fio files.FileReaderWriter, sio cmdlib.StdioRea
...
 
-	if errs := componentValidationFn(bw.AllComponents()); len(errs) > 0 {
-		return fmt.Errorf("there were one or more errors found while validating the bundle:\n%v", errs.ToAggregate())
+	bundleType := bw.Kind()
+	if bundleType == "Component" {
+		errs := validate.Component(bw.Component())
+		if len (errs) > 0 {
+			return fmt.Errorf("there were one or more errors found while validating the bundle:\n%v", errs.ToAggregate())
+		}
+	}else{   //@todo add validation for BundleBuilder, Bundle, ComponentBuilder here.
```
Kashomon: Formatting's off here. Make sure that gofmt runs automatically.

Also, could you use a proper TODO?

// TODO(BradfordMedeiros): blah

pkg/commands/validate/validate.go
```diff
@@ -45,20 +43,23 @@ func action(ctx context.Context, fio files.FileReaderWriter, sio cmdlib.StdioRea
...
+		return fmt.Errorf("Kind %v not yet supported", bundleType)
```
Kashomon: More traditional to use %q instead of %v when you have short strings

pkg/validate/validate_component.go
```diff
@@ -21,7 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	bundle "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/apis/bundle/v1alpha1"
-	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/core"
+	//"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/core"
```
Kashomon: either remove or use

pkg/validate/validate_component.go
```diff
@@ -47,34 +47,13 @@ var (
...
-		errs = append(errs, Component(c)...)
-
-		ref := c.ComponentReference()
-		p := cPath(ref)
-		if _, ok := objCollect[ref]; ok {
-			errs = append(errs, field.Duplicate(p, fmt.Sprintf("component reference %v", ref)))
-			continue
+func Components(components []*bundle.Component) field.ErrorList {
+	errs := field.ErrorList{ }
+	for _, component := range(components){
```
Kashomon: bad formatting. make sure gofmt is automatically enabled

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,249 @@
...
+        kind: Component
+        metadata:
+          name: foo-comp-1.0.2
+        spec:
+          componentName: foo-comp
+          version: 2.010.1             #invalid`,
+      expectedErrors: 1,
+      description:    "version notation is incorrect",
+    },
+    /* This is failing gives 0 errors should be 1
```
Kashomon: why?

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,249 @@
...
-  spec:
-    componentName: foo-comp
-    version: 1.0.2
-    objects:
-    - apiVersion: v1
-      kind: Pod`,
-			errSubstring: "Required value",
-		},
-	}
+func TestValidateComponents(t *testing.T) {
```
Kashomon: You removed all the err substring stuff. Why? I think that's very valuable to ensure you're actually getting the errors that you want.

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,249 @@
...
-	return nil
+func assertValidateComponentSet(componentSetString string, numErrors int, t *testing.T) {
+  componentSet, errParse := converter.FromYAMLString(componentSetString).ToComponentSet()
+  if errParse != nil {
+    t.Fatalf("parse error: %v", errParse)
+  }
+  errs := ComponentSet(componentSet)
+  numErrorsActual := len(errs)
+  if numErrorsActual != numErrors {
+    t.Fatalf("expected %v errors got %v:  %v", numErrors, numErrorsActual, errs)
```
Kashomon: Got before want: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,249 @@
...
+      componentConfig: `
+        apiVersion: bundle.gke.io/v1alpha1
+        kind: Component
+        metadata:
+          creationTimestamp: null
+        spec:
+          version: 30.0.2
+          componentName: etcd-component`,
+      expectedErrors: 0,
+      description:    "basic component validation",
```
Kashomon: description should be at the top for clarity

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,249 @@
 package validate
```
Kashomon: As a general note, you're doing a bunch of refactoring while also fixing a bug. It's much better to do these in separate PRs. most of my issues are with your test refactoring.

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,249 @@
...
-		},
+func assertValidateComponent(componentString string, numErrors int, t *testing.T) {
+  component, errParse := converter.FromYAMLString(componentString).ToComponent()
+  if errParse != nil {
+    t.Fatalf("parse error: %v", errParse)
+  }
+  errs := Component(component)
+  numErrorsActual := len(errs)
+  if numErrorsActual != numErrors {
+    t.Fatalf("expected %v errors got %v,  errors: %v", numErrors, numErrorsActual, errs)
```
Kashomon: got before want: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

pkg/build/inline_integration_test.go
```diff
@@ -82,11 +83,13 @@ func TestRealisticDataParseAndInline_Component(t *testing.T) {
 		t.Fatalf("Error calling ComponentFiles(): %v", err)
 	}
 
-	_, err = converter.FromObject(component).ToYAML()
+	yaml, err := converter.FromObject(component).ToYAML()
 	if err != nil {
 		t.Fatalf("Error converting the inlined component back into YAML: %v", err)
 	}
 
+	fmt.Println("yaml data is: ", string(yaml))
```
Kashomon: looks like a debug string

pkg/validate/validate_component_test.go
BradfordMedeiros: added it back

pkg/commands/validate/validate.go
Kashomon: Formatting is still problematic

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,257 @@
...
 
-var defaultComponentSetNoRefs = `
-apiVersion: 'bundle.gke.io/v1alpha1'
-kind: ComponentSet
-spec:
-  setName: bar-set
-  version: 1.0.2
-`
+	errString := errs.ToAggregate()
+	if errString != nil && expectedError != "" && !strings.Contains(errString.Error(), expectedError) {
```
Kashomon: Use testutil.CheckErrorSubstring

pkg/validate/validate_component_test.go
```diff
@@ -15,405 +15,257 @@
...
-components:
-- apiVersion: 'bundle.gke.io/v1alpha1'
-  metadata:
-    name: foo-comp-1.0.2
-  spec:
-    componentName: foo-comp
-    version: 1.0.2`,
-			errSubstring: "must be Component",
+			componentConfig: `
+        apiVersion: 'bundle.gke.io/v1alpha1'
```
Kashomon: Move to beginning of the line

pkg/validate/validate_component_test.go
BradfordMedeiros: done

pkg/commands/validate/validate.go
BradfordMedeiros: ran go format.
If we care about basic formatting, we really need to enforce it with a linter.

Kashomon: Yes, I agree.

Kashomon:
Can you rebase and then push that for review? there's a lot of garbage commit messages that would be good to clean up.

BradfordMedeiros:
I was just going to squash and merge.  All those commit messages will go away.

PR #179, opened 2019-02-14
Chdir to the appropriate directory when run with bazel.

Kashomon:
This changes the cwd to the test-directory when run with Bazel. This makes testing much simpler *and* compatible with bazel \o/.


johnbelamaric:
Does this work with `bazel run` or just `bazel test`? I am doing something similar in internal build code that uses bundles, but I used BUILD_WORKING_DIRECTORY.

/lgtm

Kashomon:
This will work with `bazel test`. Probably won't work with `bazel run`

PR #186, opened 2019-02-19
Add a Requirements object to specify package-time deps.

Kashomon:
This is based largely off of Go Modules.

- Require: Require components >= a specified appversion.
- Exclude: Exclude specific problematic components.

I didn't add Replace yet since it was clear how useful it would be.

Example:

```yaml
apiVersion: bundle.gke.io/v1alpha
kind: Requirements

require:

 # any foo-component will satisfy
- componentName: foo-component

 # any istio component >= 1.0 will satisfy
- componentName: istio
  appVersion: 1.0

exclude:

 # Perhaps we know that istio 1.0.4 is problematic.
- component: istio
  appVersion: 1.0.4

 # Perhaps we know that istio component version 10.3.2 is problematic.
- component: istio
  version: 10.3.2
```

Fixes: #76


BradfordMedeiros:
Also can you comment on why the version and appversion fields are two different fields.  This is a point I still find a bit confusing.  When would the version not correspond to an appversion?

Kashomon:
@BradfordMedeiros --

> I would like the config itself to communicate the idea to be communicated in the config syntax.
Uglier, but for example, renaming require to require at least would solve the problem I see.

I'm using go terminology and semantics from `go.mod`. If we were working in a void, I would agree with you, but this is based on that work. Thus, I think we should stick with `Require` and `Exclude`

> Also can you comment on why the version and appversion fields are two different fields. This is a point I still find a bit confusing. When would the version not correspond to an appversion?

This is a good question and something I've struggled with. I could not figure out a way to have a single version that does both of the things I want. 

I need to solve two problems.

1. The `version` field as it currently exists is a revision number: if you change anything about the component, you change this version. The problem this solves is that the combination of `componentName` + `version` is unique.

2. I also need version information about the underlying application (usually, one or more docker containers) for the purposes of dependencies. When component A depends on Kubernetes, the only thing it cares about (or should care about) is that Kubernetes is at `1.11.2` or whatever. It's possible to make a revision to a component without changing the underlying application version (say, by just adding a flag). 

I can't figure out a way to have one `version` field that encapsulates this. It's possible I should rename `version` to `revision` and `appVersion` to `version`. That would be a separate PR though.

Kashomon:
To summarize what I've been talking with @BradfordMedeiros, Ideally we wouldn't have both version and appVersion. I think this is possible to achieve with a virtual package system, but virtual packages also complicate things. 

This might need some more revisions, but I think this is a good start.

PR #187, opened 2019-02-19
Allow caller to define template missing key policy in patch templates

johnbelamaric:
When using this as a library, this allows application of patch templates even if not all options are known.

In particular, this can be used to generate objects that include image names that are defaulted based on the schema, which can then be extracted with find.ImageFinder.

### DIFF
--- a/pkg/options/patchtmpl/patch.go
+++ b/pkg/options/patchtmpl/patch.go
@@ -42,27 +42,28 @@ type applier struct {
 	// component objects.
 	includeTemplates bool
 
-	// Set the template missing key policy
-	missingKey string
+	// Set the template options
+	templateOpts []string
 }
 
 // NewApplier creates a new options applier instance. The filter indicates
 // keep-only options for what subsets of patches to look for.
-func NewApplier(pt *PatcherScheme, opts *filter.Options, includeTemplates bool, missingKey string) options.Applier {
-	if missingKey == "" {
-		missingKey = "error"
+func NewApplier(pt *PatcherScheme, opts *filter.Options, includeTemplates bool, templateOpts ...string) options.Applier {
+	if templateOpts == nil || len(templateOpts) == 0 {
+		templateOpts = []string{options.MissingKeyError}
 	}
+
 	return &applier{
 		scheme:           pt,
 		tmplFilter:       opts,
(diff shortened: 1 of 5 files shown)


pkg/options/patchtmpl/patch.go
```diff
@@ -41,21 +41,28 @@ type applier struct {
...
 	includeTemplates bool
+
+	// Set the template missing key policy
+	missingKey string
 }
 
 // NewApplier creates a new options applier instance. The filter indicates
 // keep-only options for what subsets of patches to look for.
-func NewApplier(pt *PatcherScheme, opts *filter.Options, includeTemplates bool) options.Applier {
+func NewApplier(pt *PatcherScheme, opts *filter.Options, includeTemplates bool, missingKey string) options.Applier {
```
Kashomon: rather than missingKey, how about having templateOptions string. So, the user would have to specify "missingkey=error".

- Alternatively, I'd recommend representing these as typed string consts.

type MissingKeyOpt string

const (
  Error MissingKeyOpt = "error"
  ...
) 

etc.

Such an enum would probably live one layer up.

johnbelamaric: Ok. I just used string so I don't have to convert back to string later. But I used constants and generalized it to options.

pkg/options/patchtmpl/patch.go
```diff
@@ -41,15 +41,23 @@ type applier struct {
...
+
+	// Set the template options
+	templateOpts []string
 }
 
 // NewApplier creates a new options applier instance. The filter indicates
 // keep-only options for what subsets of patches to look for.
-func NewApplier(pt *PatcherScheme, opts *filter.Options, includeTemplates bool) options.Applier {
+func NewApplier(pt *PatcherScheme, opts *filter.Options, includeTemplates bool, templateOpts ...string) options.Applier {
+	if templateOpts == nil || len(templateOpts) == 0 {
```
BradfordMedeiros: Why is a length 0 templateOpts invalid?

johnbelamaric: We want it to error by default, which is the previous behavior. So if the caller doesn't do anything, this will behave just like it did before.

Issue #189, opened 2019-02-19
Add version to bundlectl


Kashomon:
`bundlectl` should know it's version. Ideally, we'd have a precondition to ensure that the version of `bundlectl` doesn't stray from the release.

BradfordMedeiros:
Closed by #191

PR #191, opened 2019-02-21
Bundlectl version command

BradfordMedeiros:
Addresses  #189 
I wasn't sure what to use as the source of truth for the version information. As of now it's version.bzl
I see a version field on the root project field, but it's unclear to me if that's bundlectl version or k8s version.  Right now the source of truth in version.bzl, but we can change that to rip the info from a different place if need be w/o other changes. 

I opted to get this information statically into the binary via linking parameter, such that we don't have to manually update the code itself.    

Usage is 
~~~~
bundlectl version
~~~~
which simply prints out whatever is shoved into the version info to stdout.


BradfordMedeiros:
need to fix the merge.  the signiature for context action changed and I have to update.

BradfordMedeiros:
Okay should be good to go

cmd/bundlectl/main.go
```diff
@@ -16,18 +16,23 @@
...
+    panic("No version specified.  This should have been specified during linking")
+  }
+
+  flag.Lookup("logtostderr").Value.Set("true")
 
-	root := commands.AddCommands(context.Background(), os.Args[1:])
-	if err := root.Execute(); err != nil {
-		os.Exit(1)
-	}
+  root := commands.AddCommands(context.Background(), os.Args[1:], BUNDLECTL_VERSION)
```
Kashomon: your gofmt is not enabled again. Can you gofmt all your files and remain?

cmd/bundlectl/main.go
```diff
@@ -16,18 +16,23 @@
...
-	"os"
-
-	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/commands"
+  "context"
+  "flag"
+  "os"
+  "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/commands"
 )
 
+var BUNDLECTL_VERSION string       // This value gets statically set in go compilation.  See the bazel rules.
```
Kashomon: One thing that's not clear to me: If you `go build` or `go install` this binary, does it no longer work? That would be unfortunate in my opinion. For basically all Kubernetes tools / binaries, you can `go install` them.

BradfordMedeiros: It should. 
I'm not sure how to deal with the absolute url references to GoogleCloudPlatform to test this on my branch/repo.

It's a build flag you have to specify.   Not sure if there's a way to specify default flags.   
Once we get this on the main branch I can check it out.  If there's not an obvious way, we'll just hardcode the string.

cmd/bundlectl/version.bzl
```diff
@@ -0,0 +1 @@
+VERSION = "1"       # What should be the real source of truth for the version?  Project file, git commit, etc?
```
johnbelamaric: I don't know bazel that well. Is there a way to:

1) add in the git-commit, probably as a separate field
2) if the build is done in a dev env with uncommitted changes, then it should be flagged as dirty; e.g. if it is commit ab123c but changes are there, then this field would read "ab123c-dirty".

BradfordMedeiros: Not at my desk, but looks like there's some variables bazel references.
I'll merge it now and add that in after when I get back. 

I'm not sure we'll be able to do the dirty flag thing since I'm unsure how bazel would know if it's committed to git vs the remote server, and having it marked dirty because it's not committed locally is pointless.

Kashomon: I think I wasn't super clear before, which is my bad. I would rather we don't use bazel for the mechanics of `bundlectl version`. This repository is pretty bazel-lite, I would almost consider getting rid of bazel except that it's useful for things like prow-infra, license gathering, etc.

The disadvantage of relying on bazel, is that it makes `go install` not work. You have to build the binary with bazel. I think that would be a little sad, and also would be a bit more burden on us, because it would probably mean we would need to pre-build a bunch of binaries of different architectures.

By the way, here's what kubectl does:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/version/version.go
-> 
https://github.com/kubernetes/kubernetes/tree/master/pkg/version
-> 
https://github.com/kubernetes/kubernetes/blob/master/hack/lib/version.sh

I think we could do something similar with go-codegen, makefile, or something like hack/update-version.sh

BradfordMedeiros: The bazel flag is just the equivalent to the ldflag build option which we can pass.   It's easy to solve.  No big deal.

The only question that might be interesting is go get with the git commit hash, due to the fact that we'd have to make go get somehow able to see the git commit hash, which I'm not sure go get is actually aware of that? and at best we could in source check in one hash off, which doesn't give guarantees about how much that hash differs.

BradfordMedeiros:
Any specific items or reason we're waiting on merging this  that are unaddressed ?

Kashomon:
@BradfordMedeiros - As I mentioned, I'd like not to rely on bazel for this functionality. Let's chat wehn you're in.

BradfordMedeiros:
This code doesn't necessarily depend on bazel, as already stated.

Kashomon:
But you must compile the binary with Bazel in order for the Version to be available and if the version isn't available, then the command exits. This to me says: you must use bazel to build this binary, right?

Kashomon:
To me it would be much simpler just to have the version hard-coded in the version command. As part of the release process, we create a release branch and then we update the version of the Bundlectl to match the version of the release.

It's manual, but it ensures we don't need bazel and it's dead simple

johnbelamaric:
Perhaps pedantic, but rather than in the version command, I would create a
version package and have it as a constant in there. Or maybe create a
Version constant in the main package.

Kashomon:
> Perhaps pedantic, but rather than in the version command, I would create a version package and have it as a constant in there. Or maybe create a Version constant in the main package.

Sure, yeah. That works too.

BradfordMedeiros:
@Kashomon  good to go?

pkg/version/version.go
```diff
@@ -0,0 +1,4 @@
+package version
+
+const BundlectlVersion = "1"
```
Kashomon: Let's make this 0.10.0 or 0.10.0-alpha -- the next version we expect to release

Kashomon:
/lgtm
Let's change the version so it indicates the expected next release

BradfordMedeiros:
@Kashomon done.

PR #197, opened 2019-02-23
Re-add the copied in conversion logic from apiextenisons-apiserver

Kashomon:
Now that the cluster bundle is on K8S apimachinery >1.11, in theory we
can start using the conversion logic from 1.11 instead of this copied in
thing. Unfortunately, the internal library for apiextensions-apiserver
is going to need some work before the OpenAPI Schema validation can be
imported. So, to not block on that work, re-importing for the time
being.


Kashomon:
Actually, I don't think I need this. I fixed up the internal library \o/.

Kashomon:
Merged internal fix, so this is unnecessary

PR #199, opened 2019-02-24
 migrate from Gopkg to go mod

mikedanese:


Kashomon:
I was waiting do this because I think it can cause issues. Right now the Bundle library is intentionally fixed at Kubernetes 1.11 so that it can be imported into other libraries are fixed at, say, 1.11 via dep. With this change, it's relatively easy to introduce a dependency on, say, K8S 1.13, and thus not be able to import it into other places.

PR #200, opened 2019-02-24
 implement jsonnet applier

mikedanese:


mikedanese:
vendor/github.com/google/go-jsonnet/ast/stdast.go has a lot of lines...

Kashomon:
I'd like to wait on this until jsonnet is more supported internally so that I can still import the Cluster Bundle.

BradfordMedeiros:
Why do we want support for this jsonnet dsl?  I've never heard of it.  What's the high level intent?

Kashomon:
> Why do we want support for this jsonnet dsl? I've never heard of it. What's the high level intent?

What's the concern?

BradfordMedeiros:
The concern is I don't really understand why this is something we should be supporting, and especially at this point in development (it's just another format?).   Is there some functionality or use case that came up that this is attempting to solve ?

Kashomon:
I think `jsonnet` is pretty cool and @mikedanese has pretty jazzed about it for as long as I've known him, which is a bonus. I'm not terribly worried about the support burden, but maybe I'm overly optimistic =).

In any case, I'd like to wait until the `ObjectTemplate` lands and we ensure that jsonnet is stable enough to use

mikedanese:
> Is there some functionality or use case that came up that this is attempting to solve ?

Right now, bundles supports a text templater. Language designed for data templating (Jsonnet, but also CUE, HCL and others) will provide stronger assurances and safety over produced data, better support for abstraction, and overall better UX than go text templates. I picked jsonnet here because I use it.

> we ensure that jsonnet is stable enough to use

The jsonnet language is stable. Many teams across google are using it, possibly because it's a descendant of another widely used cluster management config language so it feels familiar. You can code search it with "file:jsonnet$" to see how it's being used. It also has some traction externally (mostly in the Kubernetes community). 

Jsonnet was originally implemented in c++. The go port is younger but is feature complete at this point. I think we would like to migrate internally usage to the go port over time. I would like to double check the go language binding API is committed to before adding dependencies on it though.

BradfordMedeiros:
@mikedanese I'm just mainly concerned about the focus of the tools and the ordering of which we do things.   This tool a configuration management lib/cli that plugs into a type of configuration pipeline, which ultimately needs to be sponsored via an actual installation mechanism.

I'm not sure this type of thing belongs in the initial implementation.  Maybe it does, but it seems like an add on feature, not core to the project.  I've been on enough projects where auxillary features are implemented before what matters that I just want to raise this as a question.

Kashomon:
Unless I'm mistaken, I think @mikedanese is super excited / enthusiastic about this. Personally, that's enough for me, at least for the moment. @BradfordMedeiros -- WDYT?

mikedanese:
I understand the concern with adding features early in a project adding to support burden. In this case, this features is fully contained to a point that was intended for extension. The upside of this feature is substantial UX improvement including the ability for component authors to move business logic out of the compiled binary and further decouple component definition from the release pipeline. I want this here or in the internal distribution, so we'll end up maintaining it somewhere.

Kashomon:
I'm down with getting this in, but let's rebase on the latest stuff.

Kashomon:
Oops. Accidentally closed.

Kashomon:
I think you should be unblocked now.

mikedanese:
Done.

pkg/options/jsonnet/jsonnet_applier.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2018 Google LLC
```
Kashomon: 2019

pkg/options/jsonnet/jsonnet_applier.go
```diff
@@ -0,0 +1,92 @@
...
+
+// NewApplier creates a new optionts applier instance.
+func NewApplier(importer jsonnet.Importer) options.Applier {
+	if importer == nil {
+		importer = disableImports{}
+	}
+	return &applier{importer}
+}
+
+func (m *applier) ApplyOptions(comp *bundle.Component, opts options.JSONOptions) (*bundle.Component, error) {
```
Kashomon: make sure to do a DeepCopy on the component.

pkg/options/jsonnet/jsonnet_applier_test.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2018 Google LLC
```
Kashomon: 2019

pkg/options/jsonnet/testdata/pod-with-imports.builder.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: bundle.gke.io/v1alpha1
```
Kashomon: How about you move all your testdata to /examples/jsonnet?

pkg/options/jsonnet/jsonnet_applier.go
mikedanese: Done

pkg/options/jsonnet/testdata/pod-with-imports.builder.yaml
mikedanese: Done

pkg/options/jsonnet/jsonnet_applier_test.go
mikedanese: Done.

pkg/options/jsonnet/jsonnet_applier.go
mikedanese: Done.

Issue #201, opened 2019-02-24
Add ObjectTemplate


Kashomon:
Right now the raw text template types assuming that you're using go templates and that you've stored the information as a config map.

That's ok, but that's not totally ideal. There's no parameter validation / defaulting. Also, it's hard to add more functionality (like more template types).

So, here's the idea: Let's create an `ObjectTemplate` type. It would look essentially the same as the `PatchTemplate`, except that it would specify a `TemplateType`

Additionally, you could have an ObjectTemplateBuilder, although it's super crucial.

Kashomon:
@mikedanese, fyi

Kashomon:
This is now done.

PR #202, opened 2019-02-24
Add ObjectTemplate type and generate codegen.

Kashomon:
Partial work on #201


mikedanese:
Is the idea that component builder and inliner will create components with these instead of config maps?

Kashomon:
> Is the idea that component builder and inliner will create components with these instead of config maps?

'Build' is a separate step from creating the objects

Generally the flow goes: `Build` (inline, make names, validation) -> `Apply` (creates the objects} -> `Filter` (get the things you want) -> `Export` (convert to just the object yamls)

We don't have a general apply. Right now we have a Patch. I think that becomes apply. And patch is one of the supported apply operations. Perhaps there's a better name than apply, though.

Kashomon:
But yes, the idea is to use this rather than ConfigMaps.

johnbelamaric:
Is this for the apply phase then? Not the build phase? So, a generalization of our current PatchTemplate?

Kashomon:
> Is this for the apply phase then? Not the build phase? So, a generalization of our current PatchTemplate?

This is for the apply phase, right. This is a generalization of the current `PatchTemplate`, yeah. However, it might make sense to keep these type separate -- I'm not sure. PatchTemplates are slightly different semantically from `ObjectTemplates`. For `PatchTemplate`, you apply options, then apply the PatchTemplates to the objects. For `ObjectTemplates` you just create objects.  That might not be a meaningful distinction though.

Kashomon:
I think this is reasonable to merge. @johnbelamaric -- I'm very curious whether think we should get rid of the PatchTemplate and replace it with ObjectTemplate of type `patch`. I can see an argument for either way.

johnbelamaric:
PatchTemplates are a two-phase apply basically, whereas Object templates are just one phase. For now let's keep them separate, but there is probably a way to generalize this concept.

Issue #155, opened 2019-02-06
[API] Add channel field to components


Kashomon:
I think components need the ability to be tagged with a 'channel' concept. So that you could, say, tag a component with `latest`. Or perhaps `alpha, beta, stable`.

Kashomon:
@johnbelamaric @justinsb 

I'm imagining this is just a 

`Channels []string` 

field on the Component / ComponentBuilder. It's possible that the channel should just be an artifact on the ComponentBuilder, and is removed after the build step.

One nice feature of the channel + building. You could move the relevant components to a `components/stable` or `components/latest` directories, as part of the component-build step.

justinsb:
I like this concept a lot.  I think we can prototype it with just a label/annotation, but if it proves valuable we should definitely make it a field on the builders.

I'm less clear on whether it should live on the component itself - I guess it would be useful because they we could know the "stable" version even if we just had the objects (i.e. were using an apiserver), but we could also encode this in the name (which would then be cleaner because name uniqueness is one of the few cross-object guarantees we make).

BradfordMedeiros:
I would have thought that the channel would imply a set of a components, and not components having a set of properties about what channel they belong to ?   Can you elaborate on the relationship between the two.

Kashomon:
First of all, I see the component as the Atom. Component Sets / Bundles probably need to be reworked or removed because they are causing endless confusion. Really ComponentSets are like ComponentDeployment objects or maybe ComponentSelector objects, which select from a bunch of components. 

But from the purpose of an cluster-component developer (let's say you work on Etcd or Istio), the thing you really care about is the Component  And you really care about the stability and release of that component.

To me, the channel represents the lifecycle of a component. The Channel allows for a slick way to 'tag' components as part of the build process.

We could call this BuildTags or similar. Maybe that would clarify things.

PR #203, opened 2019-02-25
Add build tags + an exporter for components that writes to paths.

Kashomon:
This PR adds BuildTags to the ComponentBuilder and uses them for
component exporting.

How it works:

- New repeated BuildTags field on the component-bulder object.
- This creates a new annotation when inlined.
- I reworked object / component exporting so that
  - `bundlectl export objects` just exports objects
  - `bundlectl export components` just exports components
  - `bundlectl export components --write-to=generated` will dump the
  components and the component sets into this directory. It's a
  configurable setting, but a bit fiddly.

The generated directory has the following formats (again, configurable)

`<write-to-dir>/<component-name>/<component-version>/<component-name>-<component-version>.yaml`
`<write-to-dir>/<component-name>/<build-tag>/<component-name>.yaml`
`<write-to-dir>/sets/<set-name>-<set-version>.yaml`

Example:

If 'foo' component @ version 1.0.3 has tag 'latest', then the following are generated:

```
generated/foo/1.0.3/foo-1.0.3.yaml
generated/foo/latest/foo.yaml
```

I've been fidling around with the format. Would love suggestions here.

Fixes: #155

### DIFF
--- a/pkg/apis/bundle/v1alpha1/builder_types.go
+++ b/pkg/apis/bundle/v1alpha1/builder_types.go
@@ -91,6 +91,15 @@ type ComponentBuilder struct {
 	// inline process, the data is inserted into a generated config map before
 	// being added to the objects. A ConfigMap is generated per-filegroup.
 	RawTextFiles []FileGroup `json:"rawTextFiles,omitempty"`
+
+	// BuildTags are tags to apply to the component during component-build time.
+	// The primary purpose of BuildTags is so that built-components can be
+	// exported to specific directories. It is stored as a comma-separated value
+	// in an annotation on the built-component.
+	//
+	// Build-tags should follow the same rules as Kubernetes names. See
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
+	BuildTags []string `json:"buildTags,omitempty"`
 }
 
 // FileGroup represents a collection of files. When used to create ConfigMaps
--- a/pkg/commands/export/component_export.go
+++ b/pkg/commands/export/component_export.go
@@ -1,0 +1,133 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package export
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	bundle "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/apis/bundle/v1alpha1"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/commands/cmdlib"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/converter"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/files"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/wrapper"
+	log "k8s.io/klog"
+)
+
+// pathTemplates are go-template paths to indicate how and where to generate
+// component files.  Possible values are {ComponentName, Version, BuildTag,
+// SetName}.  Only applies when the write-to flag is also set.
+var pathTemplates = []string{
+	"{{.ComponentName}}/{{.ComponentName}}-{{.Version}}.yaml",
+	"{{.ComponentName}}/{{.BuildTag}}/{{.ComponentName}}.yaml",
+	"sets/{{.SetName}}-{{.Version}}.yaml",
+}
+
+const defaultPerms = 0664
+
+// compOptions represents options flags for the export command.
+type compOptions struct {
+	// writeTo indicates the directory to write the components to. This is joined with the pa
+	writeTo string
+
+	// overwrite indicates whether to overwrite existing files.
+	overwrite bool
+
+	// exportSet determines whether to export a set for all the known components.
+	exportSet bool
+
+	// setName is an optional set name for the set of components.
+	setName string
+
+	// version is an optional version for the components
+	setVersion string
+}
+
+func componentAction(ctx context.Context, fio files.FileReaderWriter, sio cmdlib.StdioReaderWriter, o *compOptions, gopts *cmdlib.GlobalOptions) {
+	brw := cmdlib.NewBundleReaderWriter(fio, sio)
+	if err := runComponent(ctx, brw, sio, fio, o, gopts); err != nil {
+		log.Exit(err)
+	}
+}
+
+func runComponent(ctx context.Context, brw cmdlib.BundleReaderWriter, stdio cmdlib.StdioReaderWriter, fio files.FileReaderWriter, o *compOptions, gopt *cmdlib.GlobalOptions) error {
+	bw, err := brw.ReadBundleData(ctx, gopt)
+	if err != nil {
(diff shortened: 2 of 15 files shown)


Kashomon:
I think this gives a good idea of how it will work, but I'm going to rework this to be more testable

Kashomon:
Cool, I think this is ready to go now.

Kashomon:
An example of use:

```shell
bundlectl build --input-file=bundle-builder-example.yaml | bundlectl export components --write-to generated --export-set

ls -R generated

generated/data-blob:
0.1.0  latest

generated/data-blob/0.1.0:
data-blob.yaml

generated/data-blob/latest:
data-blob.yaml

generated/etcd-component:
30.0.2

generated/etcd-component/30.0.2:
etcd-component.yaml

generated/kubedns:
2.0.0

generated/kubedns/2.0.0:
kubedns.yaml

generated/kubeproxy:
2.1.0

generated/kubeproxy/2.1.0:
kubeproxy.yaml

generated/kubernetes-control-plane:
11.0.0 stable

generated/kubernetes-control-plane/11.0.0:
kubernetes-control-plane.yaml

generated/kubernetes-control-plane/stable:
kubernetes-control-plane.yaml

generated/node-config:
2.3.4

generated/node-config/2.3.4:
node-config.yaml

generated/sets:
component-set-0.1.0.yaml
```

Kashomon:
Configurable with
```
--path-templates={{.ComponentName}}/{{.Version}}/{{.ComponentName}}-{{.Version}}.yaml,{{.ComponentName}}/{{.BuildTag}}/{{.ComponentName}}.yaml,sets/{{.SetName}}-{{.Version}}.yaml,
```

pkg/apis/bundle/v1alpha1/builder_types.go
```diff
@@ -91,4 +91,13 @@ type ComponentBuilder struct {
...
 	RawTextFiles []FileGroup `json:"rawTextFiles,omitempty"`
+
+	// BuildTags are tags to apply to the component during component-build time.
+	// The primary purpose of BuildTags is so that built-components can be
+	// exported to specific directories. It is stored as a comma-separated value
+	// in an annotation on the built-component.
+	//
+	// Build-tags should follow the same rules as Kubernetes names. See
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
+	BuildTags []string `json:"buildTags,omitempty"`
```
mikedanese: Why not just use annotations?

pkg/commands/export/component_export.go
```diff
@@ -0,0 +1,133 @@
...
+		// TODO(kashomon): This is going to be a problem for mocking.
+		if _, err := os.Stat(dir); os.IsNotExist(err) {
+			os.MkdirAll(dir, os.ModePerm)
+		}
+		if !o.overwrite {
+			if _, err := os.Stat(fpath); err == nil {
+				return fmt.Errorf("file %q already exists", fpath)
+			}
+		}
+		if err := fio.WriteFile(ctx, fpath, []byte(yaml), defaultPerms); err != nil {
```
mikedanese: This should probably be atomic.

pkg/converter/exporter.go
```diff
@@ -15,46 +15,232 @@
...
+// - ComponentName: name of component
+// - Version: version of the component or component set
+// - BuildTag: build tag for the component
+// - SetName: name of the component set
+//
+// For example, here are same possible ways to craft pathTemplates
+//
+//   "{{.ComponentName}}/{{.Version}}/{{.ComponentName}}-{{.Version}}.yaml",
+//   "{{.ComponentName}}/{{.BuildTag}}/{{.ComponentName}}.yaml",
+//   "sets/{{.SetName}}-{{.Version}}.yaml",
```
mikedanese: Non-deterministic paths is really going to complicate tracking these things safely.

pkg/apis/bundle/v1alpha1/builder_types.go
Kashomon: This gets stuffed into a built-in annotation. I like the explicitness here in the Builder.

pkg/commands/export/component_export.go
Kashomon: What do you mean?

pkg/converter/exporter.go
Kashomon: I'm not following -- how is it non-deterministic?

pkg/commands/export/component_export.go
mikedanese: In the overwrite mode, errors during write can lead to corruption (e.g. is only partially completed leaving half the file written). If this is a static pod manifest, this would break a previously good configuration. I think we want to try to commit a change like this atomically like we do for projected volume files:

https://github.com/kubernetes/kubernetes/blob/81e7858ece4713e043d61d96874481134c77b086/pkg/volume/util/atomic_writer.go#L72

pkg/converter/exporter.go
mikedanese: s/non-determinisitic/non-stable/. Stability in export paths greatly simplifies the task of automatically tracking these things over time. Can you describe a scenario where this would be useful?

Kashomon: Are you suggesting that this shouldn't be configurable? I thought about doing that

mikedanese: No, just trying to understand the value of templating this vs requiring it to be a fully specified path.

pkg/commands/export/component_export.go
Kashomon: Do you feel like this is urgent and something I should do in this CL or it should be a followup? Or even an open issue that gets resolved at some point? 

It looks like making it atomic is a bit more work is why I'm asking

pkg/converter/exporter.go
Kashomon: Can you give an example? The templating was so I could support variable tags that got used as part of the path. And also because I couldn't commit to a particular format.

pkg/commands/export/component_export.go
mikedanese: No, not urgent

Kashomon: Filed https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/204

johnbelamaric: At least in my use case, there is no concurrency and any error during write would abort the whole process. So this does not seem urgent to me.

pkg/wrapper/bundlewrapper.go
```diff
@@ -149,38 +148,3 @@ func (bw *BundleWrapper) AllComponents() []*bundle.Component {
 	}
 	return nil
 }
-
-// ExportAsObjects will export a Bundle as a ComponentSet and Components,
-// or a Component as unstructured Objects.
-func (bw *BundleWrapper) ExportAsObjects() ([]*unstructured.Unstructured, error) {
```
johnbelamaric: I use this method in other tooling, I need it or an equivalent.

Kashomon: Do the methods in `exporter.go` work for you?

Kashomon:
I thought about this a little more -- I agree that the templated thing is overkill. I'm going to rework it to just hardcode the paths. It won't be configurable then. But that might be a feature rather than a bug.

Kashomon:
Ok, I hardcoded the paths in the command, but I left the templating stuff for now in the method.

pkg/converter/exporter.go
```diff
@@ -15,46 +15,232 @@
...
+func (e *Exporter) ComponentsWithPathTemplates(pathTemplates []string, set *bundle.ComponentSet) (map[string]string, error) {
+	if len(pathTemplates) == 0 {
```
BradfordMedeiros: Why is the empty set a special case?  This still should work on the empty set.

Kashomon: Good point. fixed

pkg/wrapper/bundlewrapper.go
johnbelamaric: It might work, I think I can use ComponentsAsSingleYAML then convert that to unstructured.

Kashomon:
I'm going to close this one. I don't think this is super necessary at the moment.

PR #205, opened 2019-02-26
Add component set reference type

johnbelamaric:
Adds types that are used to manage deploying and tracking status of deployment of component sets.


pkg/apis/bundle/v1alpha1/component_types.go
```diff
@@ -41,6 +41,15 @@ type ComponentSetSpec struct {
 	Components []ComponentReference `json:"components,omitempty"`
 }
 
+// ComponentSetReference provides a reference to a component set
+type ComponentSetReference struct {
```
Kashomon: Seems fine

johnbelamaric:
Ok, now it just adds the ComponentSetReference and updates the codegen.

PR #209, opened 2019-02-28
Make inline not able to be set.

### LINKED ISSUE #173: Inline is ignored for build
Inline flag is ignored for build, but not for validate of the same type, at least for some types (for example bundle builder).  

This means validation will fail at build time for the type, and not be caught in validate.

Kashomon:
Inline doesn't make sense as a flag any longer, since you can only
inline during the build phase.

Work on #173

However, more work should be done to actually clean up this code.


BradfordMedeiros:
I think this is fine to do incrementally and just take out the build flag, but I wouldn't consider it a complete removal without actually removing the flag that codes sets to handle inlining and references to it. 

https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/commands/cmdlib/global_options.go#L32

https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/commands/cmdlib/bundleio.go#L138

Kashomon:
Ok, I've removed the 'Fixed' thing

PR #211, opened 2019-03-01
Follow up pr 209 - finish removing inline option

BradfordMedeiros:
#209 Was merged in an it removed inline from cli interface, but there were some references in code that didn't get removed.

I also removed the validate url. 
We don't have urls as a concept in validation anymore.  
I didn't push on this originally (still wont push on it here, but it's incremental improvement), but it's dead code.  If we want it back for component builder validation we have source control and can get it back then.  Getting rid of it for now, but if I add a validateComponentBuilder for it's easy to track via this pr.


Kashomon:
As I mentioned, I think we should add validation for the Builders.

If you want to remove it for now, that's ok, but I'd much rather it be in a separate PR.

BradfordMedeiros:
It's dead code.  We'll add builder validation.
The bigger deal of why this needs to be removed is that the URL validation isn't actually done correctly.

URL validation and the inliner need to use the same method of resolving urls.  Under current implementation validation of url has to be kept in sync with the inliner.  That's complicated and will just cause issues eventually.

Kashomon:
>  It's dead code. We'll add builder validation.

That's fine, just trying to get the PRs to be scoped to a single idea. I know I'm less good at that -- but I'm trying to get better at that myself.

Kashomon:
/lgtm 
In the future, I'd prefer to see these as two separate PRs.

PR #213, opened 2019-03-03
Add an ObjectTemplateBuilder type.

Kashomon:
This PR adds a builder type for ObjectTemplates + inlining logic.

Additionally, this:

- Stops adding the inline annotation (TODO(self): remove completely)
- Refactors inlining to be clearer (with helper methods)
- Refactors the inlining test to check for substrs (good enough, clearer).

Partial work on #201

### DIFF
--- a/pkg/apis/bundle/v1alpha1/object_template_types.go
+++ b/pkg/apis/bundle/v1alpha1/object_template_types.go
@@ -31,11 +31,6 @@ const (
 	GoTemplate TemplateType = "go-template"
 )
 
-const (
-	// TemplateTypeIdentifier is an annotation
-	TemplateTypeIdentifier Identifier = "bundle.gke.io/template-type"
-)
-
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // ObjectTemplateBuilder contains configuration for creating ObjectTemplates.
@@ -47,8 +42,7 @@ type ObjectTemplateBuilder struct {
 	TemplateFile File `json:"templateFile,omitempty"`
 
 	// TemplateType indicates how the template should be detemplatized. By
-	// default, it assumes Go-Templates. During Build, this value is stored
-	// in the "bundle.gke.io/template-type" annotation.
+	// default, it assumes Go-Templates.
 	TemplateType TemplateType `json:"templateType,omitempty"`
 
 	// OptionsSchema is the schema for the parameters meant to be applied to
--- a/pkg/apis/bundle/v1alpha1/object_template_types.go
+++ b/pkg/apis/bundle/v1alpha1/object_template_types.go
@@ -23,8 +23,7 @@ import (
 type TemplateType string
 
 const (
-	// UndefinedTemplateType represents an undefined template type, and should
-	// default to GoTemplate.
+	// UndefinedTemplateType represents an undefined template type.
 	UndefinedTemplateType TemplateType = ""
 
 	// GoTemplate represents a go-template type.
@@ -42,7 +41,7 @@ type ObjectTemplateBuilder struct {
 	TemplateFile File `json:"templateFile,omitempty"`
 
 	// TemplateType indicates how the template should be detemplatized. By
-	// default, it assumes Go-Templates.
+	// default, it defaults to Go-Templates during build if left unspecified.
 	TemplateType TemplateType `json:"templateType,omitempty"`
 
 	// OptionsSchema is the schema for the parameters meant to be applied to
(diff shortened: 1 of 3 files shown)


pkg/apis/bundle/v1alpha1/object_template_types.go
```diff
@@ -28,21 +28,40 @@ const (
...
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	// Template is a template that creates a K8S object.
+	// Template is a template-string that creates a K8S object.
 	Template string `json:"template,omitempty"`
 
 	// TemplateType indicates how the template should be detemplatized. By
-	// default, it relies on Go-Templates.
+	// default, it assumes on Go-Templates.
```
johnbelamaric: nit: s/on //

Kashomon: clarified.

pkg/apis/bundle/v1alpha1/object_template_types.go
```diff
@@ -23,26 +23,44 @@ import (
...
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// ObjectTemplateBuilder contains configuration for creating ObjectTemplates.
+type ObjectTemplateBuilder struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// TemplateFile is references a template file
+	TemplateFile File `json:"templateFile,omitempty"`
```
mikedanese: "template" in "templateFile" and "templateType" are fairly redundant. In context, it seems pretty obvious what type and file would refer to.

pkg/apis/bundle/v1alpha1/object_template_types.go
```diff
@@ -23,26 +23,44 @@ import (
...
 
 const (
-	// UndefinedTemplateType represents an undefined template type, and should
-	// default to GoTemplate.
+	// UndefinedTemplateType represents an undefined template type.
 	UndefinedTemplateType TemplateType = ""
 
 	// GoTemplate represents a go-template type.
-	GoTemplate TemplateType = "GoTemplate"
+	GoTemplate TemplateType = "go-template"
```
mikedanese: To not pollute the package namespace, you might want to use a convention like:

```
TemplateTypeUndefined
TemplateTypeGo
TemplateTypeJsonnet
```

pkg/build/inline.go
```diff
@@ -122,12 +122,68 @@ var nonDNS = regexp.MustCompile(`[^-a-z0-9\.]`)
 // ComponentFiles reads file-references for component builder objects.
 // The returned components are copies with the file-references removed.
 func (n *Inliner) ComponentFiles(ctx context.Context, comp *bundle.ComponentBuilder) (*bundle.Component, error) {
-	name := comp.ComponentName
+	newObjs, err := n.objectFiles(ctx, comp.ObjectFiles, comp.ComponentReference())
+	if err != nil {
+		return nil, err
+	}
+
+	newObjs, err = n.objectTemplateBuilders(ctx, newObjs, comp.ComponentReference())
```
mikedanese: Does this add a file indirection to ComponentBuilder. I.e. ComponentBuilder file -> ObjectTemplateBuilder file -> template? Since ObjectTemplateBuilders are fairly concise, it would be nice to represent them inlined in ComponentBuilder.

pkg/build/inline.go
```diff
@@ -138,41 +194,95 @@ func (n *Inliner) ComponentFiles(ctx context.Context, comp *bundle.ComponentBuil
...
+			continue
+		}
+		name := obj.GetName()
+
+		builder := &bundle.ObjectTemplateBuilder{}
+		if err := converter.FromUnstructured(obj).ToObject(builder); err != nil {
+			return nil, fmt.Errorf("for component %v and object %q: %v", ref, name, err)
+		}
+
+		contents, err := n.readFile(ctx, builder.TemplateFile)
```
mikedanese: Preexisting: Relative paths in config files should probably be interpreted relative to the config file, not the current working directory.

Kashomon: Yeah, I need to fix that

pkg/build/inline.go
Kashomon: you mean like the following?

```go
type ComponentBuilder struct {
  ObjectFiles []File
  ObjectTemplates []TemplateBuilders
}

type TemplateBuilders {
  ObjectMeta
  TemplateType string
  File File
}
```

That's possible. I sort of like the modularity of the ObjectTemplateBuilders -- we largely just rely on the same tech, with the caveat there's some hooks in the inlining code. Maybe that caveat is significant enough to rethink that.

pkg/apis/bundle/v1alpha1/object_template_types.go
Kashomon: Good idea. I'll come back and do this

pkg/apis/bundle/v1alpha1/object_template_types.go
Kashomon: Good idea. I'll come back and clean this up

PR #214, opened 2019-03-03
Add a go-template applier

Kashomon:
This piggy-backs off of #213, by adding an OptionsApplier that can apply go-templates.


Kashomon:
Ok, I've rebased off of the submitted BundleTemplateBuilder PR. This should be ready to go

pkg/options/common.go
```diff
@@ -37,3 +37,30 @@ func ApplyCommon(ref bundle.ComponentReference, objs []*unstructured.Unstructure
...
+			continue
+		}
+
+		if templateType == templateKind {
+			match = append(match, obj)
+		} else {
+			notMatch = append(notMatch, obj)
+		}
+	}
+	return
```
johnbelamaric: I think it's more readable to not use the named return values / bare return.

Kashomon: Ok, that's fair. Done.

pkg/options/common.go
```diff
@@ -37,3 +37,32 @@ func ApplyCommon(ref bundle.ComponentReference, objs []*unstructured.Unstructure
 
 	return newObj, nil
 }
+
+// PartitionObjectTemplates returns all the ObjectTemplates that match a
+// specified TemplateKind and all other objects that don't match, in that
+// order.
+func PartitionObjectTemplates(allObjects []*unstructured.Unstructured, templateKind string) ([]*unstructured.Unstructured, []*unstructured.Unstructured) {
```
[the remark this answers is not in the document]
Kashomon: This is really a special case of `filter` and I want it to work the same as the Partition method there. I would just use `filter.Partition` but I have this special case where I want to partition also on the TemplateType.

Probably it should live there. 

I'm not really sure what `MatchObjectTemplatesToTemplate` means, honestly.

pkg/options/gotmpl/go_template_applier.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error executing template for component %v: %v", ref, err)
+	}
+
+	uns, err := converter.FromYAMLString(buf.String()).ToUnstructured()
+	if err != nil {
+		return nil, err
+	}
+
+	return uns, err
```
mikedanese: It's fairly common (and arguably nice) for legacy templates to produce multiple objects.  Does this require every template to be broken up to contain one object?

Kashomon: Currently, yes. Although we should support templates with multiple objects. I think this should be a follow-up PR. I'll do that next

Kashomon:
Ok, I'm merge this in to push this forward. I have some follow-up PRs that I'll be sending

1. Several object template cleanups.
  1. Add support for multiple YAML files in a single file
  2. Clean up the variable names in the types.
2. Fix the directory nonsense.
3. (Possibly): rm object template builders and have something more akin to ObjectFiles.

PR #215, opened 2019-03-04
Code quality -> Remove dead code, get rid of static gopts.

BradfordMedeiros:
Just some small touchups.   Grab bag off small stuff.

Changes:
- got rid of static gopts instance.  Instead we create gopts scoped as the command level and pass the data in as a parameter explicitly.
- removed unused createInlinerFn 
- moved opts to the file it's actually mutated in and scoped it at the function instead of statically at  package level
- commands now return the cobra command instead of mutating the root object 
- removed unused code in pkg/core (I assume we're not using this as our public api internally)
- some minor non-functional code formatting changes.


pkg/commands/cmdlib/global_options.go
```diff
@@ -26,17 +26,4 @@ type GlobalOptions struct {
...
-
-// Copy creates an options copy.
-func (g *GlobalOptions) Copy() *GlobalOptions {
-	return &GlobalOptions{
-		InputFile:    g.InputFile,
-		InputFormat:  g.InputFormat,
-		OutputFormat: g.OutputFormat,
-	}
-}
+}
```
Kashomon: +newline

Kashomon:
there's a lot of random stuff in this PR. Can you enumerate what you're changing in the CL description?

pkg/commands/validate/get_command.go
```diff
@@ -23,15 +23,14 @@ import (
 )
 
 // AddCommandsTo adds commands to a root cobra command.
-func AddCommandsTo(ctx context.Context, fio files.FileReaderWriter, sio cmdlib.StdioReaderWriter, root *cobra.Command) {
+func GetCommand(ctx context.Context, fio files.FileReaderWriter, sio cmdlib.StdioReaderWriter, gopts *cmdlib.GlobalOptions) *cobra.Command{
```
Kashomon: Why this change?

Kashomon: Make sure to run go lint, go vet -- looks like a bunch of method comments didn't get changed.

pkg/commands/build/get_command.go
```diff
@@ -22,23 +22,17 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// AddCommandsTo adds commands to a root cobra command.
-func AddCommandsTo(ctx context.Context, fio files.FileReaderWriter, sio cmdlib.StdioReaderWriter, root *cobra.Command) {
+func GetCommand(ctx context.Context, fio files.FileReaderWriter, sio cmdlib.StdioReaderWriter, gopts *cmdlib.GlobalOptions) *cobra.Command{
```
Kashomon: lost method comment

BradfordMedeiros: Can you explain why we want comments for these type of functions? 
I removed it intentionally.

pkg/commands/validate/get_command.go
BradfordMedeiros: Yea I need to remove them.

BradfordMedeiros:
@Kashomon you're right that this is a bit of a grab bag.  I  just saw some small quality of life stuff that I think were easy enough to pick up. I'll enumerate the changes in writing

pkg/commands/build/get_command.go
Kashomon: It'll ask for that in the lint because these are public functions. So, we should corresponding comments. you can run /hack/check-lint.sh or whatever it's called

pkg/commands/validate/get_command.go
BradfordMedeiros: So in detail if you want to know why:
opts was package level despite usage at function level.  This is effectively wrong.  Multiple calls to the function result in multiple references to the same static data.  same applies to gopts. 
get vs add since I was in the neighborhood.  Nice to restrict side effect to rootcmd at a single scope, just makes things easier to reason about.  :+1:

Kashomon:
Can you make sure this passes go linting?

BradfordMedeiros:
@Kashomon is the linting setup properly in this?  
when i run verify-golint i get 
~~~~
open ./pkg/...: no such file or directory
open ./cmd/...: no such file or directory
~~~~
also it doesn't seem to install golint succesfully out of the box due to path issues. 
Does this work for you?

When I hacked it up to work, I got a bunch of linting errors throughout the project for untouched files as well

Kashomon:
`./hack/verify-golint.sh` gives you those error messages? that's strange. What shell are you using?

I sent a CL to fix the outstanding lint errors in https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/pull/216

BradfordMedeiros:
bash.  had golint via apt, maybe it's different.

Kashomon:
Oh, hm. `go version go1.11.1 darwin/amd64`. I'll update to go 1.12

Kashomon:
I upgraded to go 1.12. not getting any lint errors

Kashomon:
have you been able to look into the lint issues? Would love to get this merged =)

PR #222, opened 2019-03-11
Add init command

BradfordMedeiros:
This is based off the clean up code branch.  Once that's committed I'll update this.  Right now the diff looks bigger than it needs to. 

All this does is add an initial generate command for scaffolding.  Currently I just have it setup to create a deployment and a service, you can override the name of the deployment/service, and specify the path. 

I'll add in optional test suite + patches after this.  

The question is what should be specified in scaffold, since it can really be anything. 
For example, should we just have a default service and deployment?  Maybe we allow overrides for the base container, and pull out the service to be an optional tag ?


pkg/generate/generate.go
```diff
@@ -0,0 +1,87 @@
...
+	var writeErr error
+  if err := os.Mkdir(filepath, 0777); err != nil { 
+    writeErr = err 
+    return writeErr
+  }
+
+	if err := ioutil.WriteFile(path.Join(filepath, "etcd-component-builder.yaml"), []byte(componentBuilder), 0666); err != nil {
+    writeErr = err
+  }
+	if err := ioutil.WriteFile(path.Join(filepath, "sample-deployment.yaml"), []byte(strings.Replace(sampleDeployment, "$DEPLOYMENT_NAME", name, -1)), 0666); err != nil {
```
Kashomon: How about we use real golang text templates? It's clearer and more robust than ad-hoc variable substitution, imo

(here and throughout).

pkg/generate/generate.go
```diff
@@ -0,0 +1,87 @@
...
+// Package generate scaffolds components to guide a user for component creation
+package generate
+
+import "io/ioutil"
+import "os"
+import "path"
+import "strings"
+
+const componentBuilder = `apiVersion: bundle.gke.io/v1alpha1
+kind: ComponentBuilder
```
Kashomon: Can we also add
1. an example patch
2. a test-suite test

Kashomon: If you're time limited, I'd recommend doing just the patch.

pkg/generate/generate.go
```diff
@@ -0,0 +1,87 @@
...
+  ports:
+  - name: $SERVICE_NAME
+    port: 8080
+    targetPort: 8080
+  selector:
+    app: $SERVICE_NAME
+`
+
+// Create scaffolds basic set of files to the filesystem
+func Create(filepath string, name string) error {
```
Kashomon: I'd recommend taking an options struct to Create to allow for extensibility. Or, you can do functional-options if you want to be fancy.

Dave Cheney has a nice discussion here: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Kashomon: Can you add a test?

BradfordMedeiros: Yea, I'll work on it.  Needs to be an integ test really for this type of code.

pkg/generate/generate.go
BradfordMedeiros: @Kashomon Yea let's get this functionality in an I'll merge that behavior in the morning if that's okay.

pkg/generate/generate.go
BradfordMedeiros: Probably makes sense.  I haven't used golang templates outside of ansible but it should be straight forward. 

I recognize that right not my string substitution can result in pretty broken configs since no validation on input. 

Hopefully golang templates will help constrain that, but either way stylistically makes sense.

Kashomon:
Can you rebase to remove the already-merged changes from the PR?

PR #228, opened 2019-03-11
Follow-up to ObjectTemplate code: renames, Multi-doc YAML

Kashomon:
Several followups following some comments from @mikedanese.

- renames TemplateFile to File
- renames TemplateType to Type
- enables multi-doc YAML files


pkg/options/gotmpl/go_template_applier.go
```diff
@@ -32,6 +33,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
+var multiDoc = regexp.MustCompile("---(\n|$)")
```
mikedanese: Match the beginning of the line as well.

Kashomon: `^` will cause the regex match just the beginning of entire string. 

https://play.golang.org/p/NT2W17pdTVt

mikedanese: `(\n|^)---(\n|$)` or don't use regex? Is `foo: a---` valid yaml?

Kashomon: Ah, right. Done.

mikedanese:
LGTM

Kashomon:
(note: we still use the /lgtm stuff, albeit inconsistently)

Kashomon:
Gonna merge in given the LGTM.

PR #230, opened 2019-03-11
Make a copy of the component before doing transformations.

Kashomon:


johnbelamaric:
Does this fix a bug? Maybe a test reproducing the bug and showing this fixes it?

Kashomon:
This fixes a bug that I encountered internally. I.e., -- I had a test and modified a component, but I was reusing the same base component (and was getting errors). I could make a test

Kashomon:
Added a test

pkg/options/gotmpl/go_template_applier.go
```diff
@@ -51,6 +51,9 @@ func NewApplier() options.Applier {
 // component's object list. The original ConfigMap is not included in the final
 // component.
 func (m *applier) ApplyOptions(comp *bundle.Component, opts options.JSONOptions) (*bundle.Component, error) {
+	// Make a copy to avoid confusing behavior.
+	comp = comp.DeepCopy()
```
BradfordMedeiros: The two appliers now have different behavior but similar interfaces, see:
https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/options/patchtmpl/patch.go#L75
If you update this fella, you should update the equivalent ones to also create deep copies. 

the variable shadowing is a bit confusing, I'd just create a separate reference but stylistic thing so whatever.

Kashomon: the `:=` would have been variable shadowing. This is just replacing the reference.

Added the DeepCopy to patch. the `rawtexttmpl` and `simpletemplate` already have it.

PR #231, opened 2019-03-11
Init command

BradfordMedeiros:
Made new pr.  rebasing had was messy, easier to just open new pr. 
Added templates instead of the simple string replace.  

Still working on the extra patching and test suite integration on this pr and that can either be in this or a separate pr, doesn't matter from my perspective.

### DIFF
--- a/pkg/commands/generate/get_command.go
+++ b/pkg/commands/generate/get_command.go
@@ -1,0 +1,43 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package generate contains the generate command.
+package generate
+
+import (
+	"os"
+	"github.com/spf13/cobra"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/generate"
+)
+
+// GetCommand generate placeholder components
+func GetCommand() *cobra.Command {
+	var filepath string
+	var componentName string
+	cmd := &cobra.Command{
+		Use:   "generate",
+		Short: "Generate placeholder components",
+		Long:  "Generate placeholder components",
+		Run: func(cmd *cobra.Command, _ []string) {
+			if filepath == "" {
+				os.Stderr.WriteString("Filepath not specified")
+				os.Exit(1)
+			}
+			generate.Create(filepath, componentName)
+		},
+	}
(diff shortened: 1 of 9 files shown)


pkg/commands/generate/get_command.go
```diff
@@ -0,0 +1,43 @@
...
+func GetCommand() *cobra.Command {
+	var filepath string
+	var componentName string
+	cmd := &cobra.Command{
+		Use:   "generate",
+		Short: "Generate placeholder components",
+		Long:  "Generate placeholder components",
+		Run: func(cmd *cobra.Command, _ []string) {
+			if filepath == "" {
+				os.Stderr.WriteString("Filepath not specified")
```
Kashomon: log.Exit("....")

pkg/commands/generate/get_command.go
```diff
@@ -0,0 +1,43 @@
...
+				os.Exit(1)
+			}
+			generate.Create(filepath, componentName)
+		},
+	}
+	cmd.Flags().StringVar(&filepath, "path", "", "filepath to create component")
```
Kashomon: how about "write-to" to make this consistent with the `export` command? (well, when that lands).

I'd also mention that this is required

pkg/commands/generate/get_command.go
```diff
@@ -0,0 +1,43 @@
...
+package generate
+
+import (
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/generate"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+// GetCommand generate placeholder components
+func GetCommand() *cobra.Command {
```
Kashomon: This should receive dependencies (file-writer).

pkg/generate/generate.go
```diff
@@ -0,0 +1,138 @@
...
+      type: string
+      pattern: '^[a-z0-9-]+$'
+`
+const patchOptions = `# Options for applying to the component via patch-templates
+namespace: foo-namespace
+buildLabel: test-build
+`
+
+// Create scaffolds basic set of files to the filesystem
+func Create(filepath string, name string) error {
```
Kashomon: I'd just return map[string]contents and write out the data in the Command. Limits the I/O to the commands and makes this function easier to test.

Alternatively, pass in the filereaderwriter

pkg/generate/generate.go
```diff
@@ -0,0 +1,138 @@
...
+		goto handleError
+	}
+	if err := ioutil.WriteFile(path.Join(filepath, "sample-options.yaml"), []byte(patchOptions), 0666); err != nil {
+		writeErr = err
+		goto handleError
+	}
+
+handleError:
+	if writeErr != nil {
+		os.Remove(filepath)
```
Kashomon: I think it's fine to write partial results. Also, the go-to stuff is pretty weird

pkg/commands/generate/get_command.go
BradfordMedeiros: Not going to use file writer in this code unless we remove references to bundle in filewriter. 
No reason for this code to depend on the bundle implementation.

pkg/generate/generate.go
BradfordMedeiros: okay.  I think in practice that should work well.

Kashomon:
/lgtm 
This is a good start. we can iterate on this.

Issue #227, opened 2019-03-11
Fix Pathing for building: Add relative paths back


Kashomon:
Right now, build only works based on absolute paths (given some root directory). Although workable, it's pretty confusing. It would be much clearer to support that world, while also supporting relative paths (which everybody will expect).

Kashomon:
@mikedanese @BradfordMedeiros fyi.

PR #233, opened 2019-03-12
Add relative-path support back.

Kashomon:
This PR adds relative path support back. You can now specify relative
paths for any of the file URLs. File-paths will be rewritten only if the
child-file path (e.g., an object in a component) is a relative path.

Fixes: #227

### DIFF
--- a/examples/cluster/etcd/etcd-component.yaml
+++ b/examples/cluster/etcd/etcd-component.yaml
@@ -17,4 +17,4 @@ kind: ComponentBuilder
 componentName: etcd-component
 version: 30.0.2
 objectFiles:
-- url: file:///etcd/etcd-server.yaml
+- url: ./etcd-server.yaml
--- a/pkg/build/path_rewriter.go
+++ b/pkg/build/path_rewriter.go
@@ -22,18 +22,21 @@ import (
 // RelativePathRewriter rewrites paths when the paths are relative paths.
 type RelativePathRewriter struct{}
 
-// RewritePath rewrites file paths if the path is relative, from the
+// MakeAbs rewrites file paths if the path is relative, from the
 // original object path, to an object path that's based on a parent component
-// path.
+// path. Note that the parent path must be an absolute path.
 //
 // For example, if the component path is foo/bar/biff.yaml and the object path
 // is zed/fred.yaml, the object will be rewritten as foo/bar/zed/fred.yaml
-func (rw *RelativePathRewriter) RewriteURL(parent, obj *url.URL) *url.URL {
+func (rw *RelativePathRewriter) MakeAbs(parent, obj *url.URL) *url.URL {
 	if parent == nil || obj == nil {
 		return obj
 	}
 	if parent.Scheme != "file" && parent.Scheme != "" {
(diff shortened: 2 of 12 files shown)


pkg/build/path_rewriter.go
```diff
@@ -0,0 +1,50 @@
...
+// RelativePathRewriter rewrites paths when the paths are relative paths.
+type RelativePathRewriter struct{}
+
+// RewritePath rewrites file paths if the path is relative, from the
+// original object path, to an object path that's based on a parent component
+// path.
+//
+// For example, if the component path is foo/bar/biff.yaml and the object path
+// is zed/fred.yaml, the object will be rewritten as foo/bar/zed/fred.yaml
+func (rw *RelativePathRewriter) RewriteURL(parent, obj *url.URL) *url.URL {
```
mikedanese: Myabe call this Abs() or MakeAbsolute()

pkg/build/path_rewriter.go
```diff
@@ -0,0 +1,50 @@
...
+
+package build
+
+import (
+	"net/url"
+	"path/filepath"
+)
+
+// RelativePathRewriter rewrites paths when the paths are relative paths.
+type RelativePathRewriter struct{}
```
mikedanese: Are you thinking about adding any other rewriters? What is the motivation for the struct vs a function?

pkg/build/path_rewriter.go
```diff
@@ -0,0 +1,50 @@
...
+		// Only file schemes are supported.
+		return obj
+	}
+	if filepath.IsAbs(obj.Path) {
+		return obj
+	}
+	return &url.URL{
+		Scheme: parent.Scheme,
+		Host:   parent.Host,
+		Path:   filepath.Join(filepath.Dir(parent.Path), obj.Path),
```
mikedanese: Do you want to filepath.Clean this as well? Do you want to add any limits like disallowing traversal out of the workspace?

examples/cluster/etcd/etcd-component.yaml
```diff
@@ -17,4 +17,4 @@ kind: ComponentBuilder
 componentName: etcd-component
 version: 30.0.2
 objectFiles:
-- url: file:///etcd/etcd-server.yaml
+- url: ./etcd-server.yaml
```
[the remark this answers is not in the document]
Kashomon: This uses url.Parse(...). Whatever that accepts is valid

pkg/build/path_rewriter.go
Kashomon: I thought that I might want different rewriters. /shrug. I don't have a strong opinion.

pkg/build/path_rewriter.go
Kashomon: Done.

Kashomon: I don't know how to disallow traversals out of the workspace. =/

pkg/build/path_rewriter.go
Kashomon: I did MakeAbs(...)

mikedanese:
LGTM

PR #237, opened 2019-03-12
Some cleanups for the generate command.

Kashomon:


pkg/commands/generate/get_command.go
```diff
@@ -18,25 +18,24 @@ package generate
...
 func GetCommand() *cobra.Command {
 	var filepath string
 	var componentName string
+
 	cmd := &cobra.Command{
 		Use:   "generate",
 		Short: "Generate placeholder components",
 		Long:  "Generate placeholder components",
 		Run: func(cmd *cobra.Command, _ []string) {
-			if filepath == "" {
```
BradfordMedeiros: Shouldn't we check this?

pkg/generate/generate.go
```diff
@@ -15,15 +15,19 @@
...
+	"os"
+	"path"
+	"text/template"
+
+	log "k8s.io/klog"
+)
 
 const componentBuilder = `apiVersion: bundle.gke.io/v1alpha1
 kind: ComponentBuilder
-componentName: etcd-component
```
BradfordMedeiros: woopsie good catch.

pkg/generate/generate.go
```diff
@@ -102,24 +108,26 @@ func Create(filepath string, name string) error {
...
+	var compText bytes.Buffer
+	compTmpl.Execute(&compText, replacement)
+
+	if err := ioutil.WriteFile(path.Join(dir, replacement.Name+"-builder.yaml"), compText.Bytes(), 0666); err != nil {
+		log.Exit(err)
 	}
-	if err := ioutil.WriteFile(path.Join(filepath, "sample-deployment.yaml"), deploymentText.Bytes(), 0666); err != nil {
-		writeErr = err
+	if err := ioutil.WriteFile(path.Join(dir, "sample-deployment.yaml"), deploymentText.Bytes(), 0666); err != nil {
+		log.Exit(err)
```
BradfordMedeiros: with log.exit you're making assumptions about how this is used.  Ideally you should allow the calling context retain the ability to decide how to handle the error.

pkg/commands/generate/get_command.go
Kashomon: I changed it to rely on the CWD by default (by essentially allowing an empty filepath).

pkg/generate/generate.go
Kashomon: Good point. I'll just return the error.

Kashomon: I'll do that in a follow-up. I've got some more cleanups I was thinking about.

Issue #234, opened 2019-03-12
Virtual Package Support


Kashomon:
Currently components have AppVersion and Version. I think we could do away with AppVersion by having virtual components. These would be components with no objects but a requirements that just specified names of components. Then, the child-components could require specific versions of the parent.

E.g., 

```
Component: Kubernetes
Version: 1.11
Requires
- ApiServer
- KubeControllerManager

Component: ApiServer
Version: 10.2.3
Requires:
- Kubernetes, >1.11

Component: KubeControllerManager
Version: 10.2.3
Requires:
- Kubernetes, >1.10
```

Kashomon:
I added this with the dependency solver via Visibility

PR #238, opened 2019-03-13
Remove AppVersion from components, excludes from Requirements.

Kashomon:
I think we can just get away with Requires for the time being and worry
about excludes later.

Additionally, the virtual package support mentioned in #234 should
allow us to get all the benefits that we were getting from AppVersion.


Kashomon:
I'm not sure this is such a good idea. We might still want `appVersion`. I'm going to close this and think on it for now.

Issue #126, opened 2019-01-15
Testing: Add E2E tests for the bundle CLI (bundlectl)


artasparks:
I'd love to have real E2E tests for the Bundle CLI which build the bundlectl binary and use it to process bundle / component files.

PR #242, opened 2019-03-14
Add a intgration test framework for bundlectl commands.

Kashomon:
This PR adds an integration test framework for commands by mocking out
stdio, fileio, and exiting. Additionally I refactored the bundleio test
slightly to use the new fakes and I used the integration test framework
for the easiest command: `bundlectl version`.

Still to do:

- Use the CmdIO struct everywhere.
- Make more integration tests.

Partial work on #126

### DIFF
--- a/pkg/apis/bundle/v1alpha1/object_template_types.go
+++ b/pkg/apis/bundle/v1alpha1/object_template_types.go
@@ -26,7 +26,7 @@ const (
 	// TemplateTypeUndefined represents an undefined template type.
 	TemplateTypeUndefined TemplateType = ""
 
-	// TemplateGo represents a go-template, which is assumed to be YAML.
+	// TemplateTypeGo represents a go-template, which is assumed to be YAML.
 	TemplateTypeGo TemplateType = "go-template"
 )
 
--- a/pkg/commands/cmdlib/BUILD.bazel
+++ b/pkg/commands/cmdlib/BUILD.bazel
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "bundleio.go",
+        "cmdio.go",
         "doc.go",
         "global_options.go",
     ],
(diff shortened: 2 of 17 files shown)


pkg/commands/version/get_command.go
```diff
@@ -16,20 +16,22 @@
...
-	"fmt"
-
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/commands/cmdlib"
 	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/version"
 	"github.com/spf13/cobra"
 )
 
 // GetCommand returns the version command.
-func GetCommand() *cobra.Command {
+func GetCommand(cio *cmdlib.CmdIO) *cobra.Command {
```
BradfordMedeiros: I just want to point this out, and I'm not blocking you on this, I just  would personally avoid doing this, and it's a good simple low impact example of a problem I often see:

I'll explain:
By doing this, you are now making the version package depend upon cmdlib. 
Cmdlib target depends on 
~~~~
        "//pkg/apis/bundle/v1alpha1:go_default_library",
        "//pkg/build:go_default_library",
        "//pkg/converter:go_default_library",
        "//pkg/files:go_default_library",
        "//pkg/options:go_default_library",
        "//pkg/wrapper:go_default_library",
~~~~

This isn't ideal.  This couples the code together.

Instead, if you want this same behavior, and you want to use this,  have getcommand take interface that are defined in the version package, and bind the implementation elsewhere. 
For example, you can easily, easily define a writeTo via an arg writeTo(stuffToWrite string), or an onWriteComplete()  local to this module.

You can then bind it in cmdlib. 

In general I encourage you to not directly take dependencies on code that you do not need to.  
You end up making this piece of code bound to the implementation of other modules (because it directly shares code with them)

Kashomon: I could move cmdlib into `cmdio` and `bundleio` to break it up. Would that help from your perspective?

Kashomon:
I'm ok with adding a full E2E test suite, but I think it's easier to do E2E tests for most things with this framework rather than shell. Shell is a pain to do testing with.

PR #243, opened 2019-03-14
simple example test using version

BradfordMedeiros:
Added a simple example test for full integration testing for the bundlectl command.
I added a simple, simple test which simply verifies that the version command exists. 

So just for clarity, and maybe this should be documented since it's non-obvious, but these sh_tests basically pass or fail depending on the exit code.

I plan on adding in a variety of tests today, but I just want to get this in asap as an example test.


BradfordMedeiros:
Also one thought @Kashomon --> would you be opposed to me adding a top level makefile? 
I personally like to standardize on make build, make test, make publish, make install, etc 

I'm using bazel as a dependency to test.  I think that's okay. 
Alternatively we could do this via a make script.

In any case, the test code is thankfully not too specific to bazel.

Kashomon:
I'd prefer to use Bazel over make, since that has nice integration with Prow. Also, bazel has other benefits too (blaze query, for example) that are nice.

Did you see https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/pull/242? Is this a response to that CL?

Kashomon:
If we're going to do Shell, I would prefer we follow the GBash style guide https://google.github.io/styleguide/shell.xml and probably also use the GBash test runner

Kashomon:
oops. didn't mean to close

BradfordMedeiros:
@Kashomon no this was miscoordinated if anything, not supposed to be in response to the other pr.

Kashomon:
No worries

cmd/bundlectl/integration-tests/version.sh
```diff
@@ -0,0 +1,3 @@
+
+bundlectl="./$1"
```
Kashomon: + #!/bin/bash
+ copyright header 

set -o errexit
set -o nounset
set -o pipefail

cmd/bundlectl/integration-tests/BUILD
```diff
@@ -0,0 +1,9 @@
+sh_test(
```
Kashomon: Can we use the directory `integration_tests` or `integration`? `blaze` (internal) doesn't play nice with foo-bar directories, as I recall

cmd/bundlectl/integration-tests/version.sh
Kashomon: this is specifically what I meant about following the bash style guide.

BradfordMedeiros:
@Kashomon is this okay to merge?  I think it's valuable to have full E2E tests for cmd line utiltiies.

Kashomon:
/lgtm
Sure. I would strongly prefer most of our integration tests be written in Go using the thing I submitted, but it's useful to have a couple smoke tests like this. Doesn't need to be very many (maybe version is enough?)

BradfordMedeiros:
@Kashomon my intent is to have full end to end tests.  Basically move the example directory into tests. 
Consider that most of these tests will be having input files, and comparing them to a resultant set of output files.  Bash is excellent for that.

Kashomon:
I don't think shell tests give more coverage than the go integration tests

PR #246, opened 2019-03-18
Add selector to patch template types

johnbelamaric:


pkg/apis/bundle/v1alpha1/patch_template_types.go
```diff
@@ -19,6 +19,30 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// ObjectSelector is used for identifying Objects on which to apply the patch template.
+type ObjectSelector struct {
```
Kashomon: +json annotations

pkg/apis/bundle/v1alpha1/patch_template_types.go
```diff
@@ -19,6 +19,30 @@ import (
...
+        // Kinds represent the Kinds to filter on.
+        Kinds []string
+
+        // Names represent the names to filter on. For objects, this is the
+        // metadata.name field. For components, this is the ComponentName.
```
Kashomon: rm `For components...`

pkg/apis/bundle/v1alpha1/patch_template_types.go
```diff
@@ -19,6 +19,30 @@ import (
...
+        Annotations map[string]string
+
+        // Labels contain key/value pairs to filter on. An empty string value matches
+        // all label-values for a particular key.
+        Labels map[string]string
+
+        // Namespaces to filter on.
+        Namespaces []string
+
+        // Invert the match
```
Kashomon: By default does this match things or filter things? I assume the former; would be good to call out explicitly

pkg/apis/bundle/v1alpha1/patch_template_types.go
```diff
@@ -19,6 +19,35 @@ import (
...
+	Namespaces []string `json:"namespaces,omitempty"`
+
+	// NamespacedOnly matches only if the type of object is namespaced (i.e.,
+	// is not a cluster-wide resource).
+	NamespacedOnly *bool `json:"namespacedOnly,omitempty"`
+
+	// NegativeMatch invert the match. By default, the ObjectSelector will include
+	// objects matching all of the criteria above. This flag indicates that objects
+	// NOT matching the criteria should be included instead.
+	NegativeMatch *bool `json:"negativeMatch,omitempty"`
```
Kashomon: I don't think this needs to be a bool-pointer. Since not specifying this is precisely equivalent to false

johnbelamaric:
API conventions require it: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

In fact we should go through and +optional all the optional fields, but that is for the open api schema only which we are not generating at this time.

PR #257, opened 2019-03-24
Add support for the ObjectSelector.

Kashomon:
This is a significant (and breaking) change to the way patching works.

- Patching now relies on the ObjectSelector to determine which objects
  to apply patches to
- If a Patch has 'kind' or 'apiVersion' fields, these are removed and
  added to the ObjectSelector. This allows supporting the existing
  workflow, although matching on 'metadata.name' via the Template is no
  longer supported.
- Change NegativeMatch to InvertMatch. I like this phrasing a bit better.
- Removed the NamespacedOnly option. It's not clear to me that this
  should be a selector option. Maybe it should just be an option on the
  Patch. In any case, I think it's prudent to add support for this in a
  follow-up PR.
- Change the KeepOnly flag and Filter Option to InvertMatch for
  consistency.
- Rewrite the Filter library to support Select and Filter methods.
  Select(+InvertMatch) == Filter(), but I think it makes the usage of
  the library significantly clearer (I always got confused which was
  which).


johnbelamaric:
Awesome, looks good.

/lgtm

Issue #253, opened 2019-03-21
Promote to v1beta1


Kashomon:
It's getting to the point where breaking changes are painful and we're not making many API changes. As such, I'd like to consider promoting the API to v1beta1. I'll try to put together a list of blockers.

PR #261, opened 2019-04-12
Promote the Component to v1beta1

Kashomon:
I think this is the only type I'm confident is stable enough to be
promoted. ComponentSets might be close enough also.

After this, need to clean up existing references to v1alpha1 components. Will be a bit tricky because of the Bundle type. Might need to get rid of that first.

Partial work on #253


Kashomon:
abandoning this because I'm removing appVerison

Issue #262, opened 2019-04-15
bundle should handle CustomResources


esert-g:
We want to use/patch CustomResources in bundles, but we can't do that because patcher doesn't know about CRD schemes (see [patcher scheme](https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/1f578d4e1a6165c27ca7742c600396d2e78e298a/pkg/options/patchtmpl/scheme.go#L44-L52)). Easiest way to address this would be to bring back generic json patcher as an option to patch templates.

PR #264, opened 2019-04-15
Use patch type to apply JSON patches

esert-g:
This PR adds code for using JSON patches when `PatchTemplate`'s `patchType` field is set to `JSONPatch`.

This is the remaining work for https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/262


Kashomon:
I wonder if we should call it JSONPatch6902 or if this is sufficient. WDYT?

Kashomon:
I'm tending towards calling this patch type just as you've described it.

esert-g:
I think JSONPatch is good enough. We can always add a second more descriptive JSON patch type if there is another in-demand JSON patching method in future.

pkg/options/patchtmpl/patch.go
```diff
@@ -234,28 +252,38 @@ func objectApplier(scheme *PatcherScheme, patches []*parsedPatch) options.ObjHan
...
 			if !canApplyPatch(pat, obj) {
 				continue
 			}
 
-			newObjJSON, err := strategicpatch.StrategicMergeMapPatchUsingLookupPatchMeta(objJSON, pat.jsonMap, objSchema)
-			if err != nil {
-				return nil, fmt.Errorf("while applying patch\n%sto \n%s: %v", pat.raw, objJSON, err)
+			var newObjJSON map[string]interface{}
+			if pat.patchType == bundle.JSONPatch {
+				if objByt, err := converter.FromObject(objJSON).ToJSON(); err != nil {
```
Kashomon: I think it would be clearer if you didn't shadow-variables here. objByt is defined above

pkg/options/patchtmpl/patch.go
```diff
@@ -234,28 +252,38 @@ func objectApplier(scheme *PatcherScheme, patches []*parsedPatch) options.ObjHan
...
 		for _, pat := range patches {
 			if !canApplyPatch(pat, obj) {
 				continue
 			}
 
-			newObjJSON, err := strategicpatch.StrategicMergeMapPatchUsingLookupPatchMeta(objJSON, pat.jsonMap, objSchema)
-			if err != nil {
-				return nil, fmt.Errorf("while applying patch\n%sto \n%s: %v", pat.raw, objJSON, err)
+			var newObjJSON map[string]interface{}
+			if pat.patchType == bundle.JSONPatch {
```
Kashomon: Maybe this should be a switch too? I don't have a strong opinion.

pkg/options/patchtmpl/patch.go
```diff
@@ -138,6 +142,19 @@ func (a *applier) makePatches(ptObjs, objs []*unstructured.Unstructured, opts op
 	// Next, de-templatize the templates.
 	var patches []*parsedPatch
 	for j, pto := range pts {
+		patchType := bundle.PatchType(pto.PatchType)
+		switch patchType {
+		case bundle.StrategicMergePatch, bundle.JSONPatch:
+			// good types
```
Kashomon: s/good/known types

pkg/apis/bundle/v1alpha1/patch_template_types.go
```diff
@@ -52,6 +52,9 @@ const (
 	// Merge Patch to apply patches. See more at:
 	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md
 	StrategicMergePatch PatchType = "StrategicMergePatch"
+
+	// Use JSON patch for non-kubernetes types, such as custom resources.
```
Kashomon: I would change tho comment to be something like "JSONPatch relies on RFC6902 JSON Patching to apply patches and is useful for when objects are not rendered in the patcher runtime schema."

pkg/options/patchtmpl/patch.go
esert-g: done

pkg/options/patchtmpl/patch.go
esert-g: done

pkg/options/patchtmpl/patch.go
esert-g: done

pkg/apis/bundle/v1alpha1/patch_template_types.go
esert-g: done

pkg/apis/bundle/v1alpha1/patch_template_types.go
```diff
@@ -52,6 +52,10 @@ const (
 	// Merge Patch to apply patches. See more at:
 	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md
 	StrategicMergePatch PatchType = "StrategicMergePatch"
+
+	// JSONPatch relies on RFC6902 JSON Patching to apply patches and is useful
+	// for when objects are not rendered in the patcher runtime schema.
```
Kashomon: s/objects/object types
s/rendered/available

Sorry, my bad

esert-g: done

PR #267, opened 2019-04-29
Add a dependency solver

Kashomon:
This PR:

- Removes AppVersion and replaces with visibility in Requirements. This
  makes it much easier to reason about dependencies.
- Creates a new latest-version-perfering dependency solver
- Adds support for Match/Exclude labels during dependency selection
- Adds many tests to ensure everything works.

### DIFF
--- a/pkg/deps/collate.go
+++ b/pkg/deps/collate.go
@@ -22,19 +22,6 @@ import (
 	"github.com/blang/semver"
 )
 
-// searchOpts are options for searching for latest or previous versions. For a
-// more detailed discussion, see the ResolveOptions.
-type searchOpts struct {
-	match          map[string][]string
-	matchIfPresent map[string][]string
-	exclude        map[string][]string
-}
-
-// String returns the stringified search options.
-func (s *searchOpts) String() string {
-	return fmt.Sprintf("{%+v, %+v}", s.match, s.exclude)
-}
-
 // sortedVersions contain all the component versions for a single component,
 // sorted by version. It should be considered immutable and must not be changed
 // once created.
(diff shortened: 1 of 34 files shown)


pkg/deps/latest.go
```diff
@@ -0,0 +1,244 @@
...
+				return nil, err
+			}
+		}
+	}
+	return curNode.children, nil
+}
+
+// processPicked checks to make sure an already picked component works with the
+// existing constraints
+func (r *graphBuilder) processPicked(dep *requestedDep, curNode, picked *node) (bool, error) {
```
mamaso: Remove the receiver as it is unused.

pkg/deps/collate.go
```diff
@@ -0,0 +1,188 @@
...
+func (s *sortedVersions) useComponent(m *depMeta, opts *searchOpts) bool {
+	if opts == nil {
+		return true
+	}
+
+	match := true
+	for k, vals := range opts.match {
+		matchesOne := false
+		for i := range vals {
+			if m.annotations[k] == vals[i] {
```
BradfordMedeiros: What happens if m.annotations is length 1 and opts.match is length 5? 
Seems like that errors, why is that invalid inputs?

pkg/deps/latest.go
```diff
@@ -0,0 +1,244 @@
...
+	if picked.meta.version.GTE(dep.version) {
+		// It works!
+		return true, nil
+	}
+	if dep.version.GT(picked.maxOfMinReq) {
+		picked.maxOfMinReq = dep.version
```
mamaso: Can you add a comment to explain `maxOfMinReq` more clearly? That's the field I have the hardest time understanding.

pkg/deps/collate.go
Kashomon: Oh, good point. this is wrong. It should be matching all

pkg/deps/latest.go
```diff
@@ -0,0 +1,244 @@
...
+func (r *graphBuilder) pickLatest(dep *requestedDep, curNode *node) (*node, error) {
+	sorted, ok := r.componentVersions[dep.componentName]
+	if !ok {
+		return nil, fmt.Errorf("could not find sorted component versions for component name %q", dep.componentName)
+	}
+	latest, err := sorted.latest(r.searchOpts)
+	if err != nil {
+		return nil, err
+	}
+	if latest.version.LT(dep.version) {
```
mamaso: Move this control logic to the calling function - I'd prefer not to communicate state transitions via nil pointer

pkg/deps/resolve.go
```diff
@@ -0,0 +1,167 @@
...
+	// be present.
+	//
+	// In other words, this is a logical AND operation of the passed in Annotation and
+	// the component Annotation.
+	Match map[string][]string
+
+	// MatchIfPresent are component annotations that are analyzed if they are
+	// present on the component. This is useful for synchronizing values between
+	// the universe of components and the caller.
+	MatchIfPresent map[string][]string
```
mamaso: `MatchIfPresent` could use an example in the description of matching with and without present annotation

pkg/deps/resolve.go
```diff
@@ -0,0 +1,167 @@
...
+	Exclude map[string][]string
+}
+
+// ResolveLatest resolves dependencies for several components, returning
+// the components references that correspond to that version selection.
+//
+// These components must exist within the Resolver's set of components. If a
+// version is not specified for a component reference, the latest version is
+// used.
+func (r *Resolver) ResolveLatest(refs []bundle.ComponentReference, opts *ResolveOptions) ([]bundle.ComponentReference, error) {
```
mamaso: 'Latest' is a bit of a misnomer, as some versions may be downgraded from latest.

pkg/deps/resolve.go
```diff
@@ -0,0 +1,167 @@
...
+
+// HasComponent indicates whether a component is known by the resolver.
+func (r *Resolver) HasComponent(ref bundle.ComponentReference) bool {
+	return r.componentLookup[ref] != nil
+}
+
+// Add adds components to the resolver, returning a new resolver or
+// an error if it fails. If one of the new components provided has the same
+// name/version, it overwrites the existing component map.
+func (r *Resolver) Add(newComps []*bundle.Component) (*Resolver, error) {
```
BradfordMedeiros: Do we need Add?

pkg/deps/latest.go
```diff
@@ -0,0 +1,244 @@
...
+
+// tryDowngrade tries to downgrade the current node version, in-place.
+//
+// This is done in the case where there was some, possibly recoverable
+// incomptibility. Because we always select the latest version of dependencies,
+// we don't have the option to upgrade the picked version. Instead, we can try
+// to downgrade current node and start the dependency process over for the
+// current node.
+func (r *graphBuilder) tryDowngrade(curNode *node) error {
+	if curNode.fixed {
```
BradfordMedeiros: It's unobvious within this function what the affect of setting currNode.children to null and meta to previous would do. Those are not concepts that actually are from this function.  Instead yoru depending upon class level data and how this code was called. 

Consider returning data instead.  TryGetLowerVersion which returns  lower version if possible.  Don't mutate anything in here, and only modify the data structures from the calling piece of code.  

Notice your comments " Clear out the children and try again".  That's not what the code does.  That's only what the code does in the context of the other code that uses this code has the affect of doing.   I encourage you to apply the mindset of returning data and only mutating it as close to the actual relevant data structure manipulating code as much as possible.

pkg/deps/meta.go
```diff
@@ -0,0 +1,177 @@
...
+func (m *depMeta) versionStr() string { return m.version.String() }
+
+// ref makes a component reference.
+func (m *depMeta) ref() bundle.ComponentReference {
+	return bundle.ComponentReference{m.componentName, m.version.String()}
+}
+
+// visibleTo returns whether this component is visible to another component.
+func (m *depMeta) visibleTo(other *depMeta) bool {
+	if len(m.visibility) == 0 || m.visibility["@private"] {
```
BradfordMedeiros: still not sold on the idea that we need public/private labels.  Still see this as complexity that we should avoid for now.

pkg/deps/latest.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+			curNode.meta = previous
+		}
+	}
+	return candidateChildren, nil
+}
+
+// processPicked checks to make sure an already picked component works with the
+// existing constraints
+func processPicked(dep *requestedDep, curNode, picked *node) (bool, error) {
```
mamaso: error is never used, please remove

PR #268, opened 2019-04-30
Basic benchmarks

BradfordMedeiros:
These are only basic tests, but I'm getting 
BenchmarkBuildAndInline_Component-12    	     500	   2733579 ns/op
BenchmarkBuildAndPatch_Component-12     	    2000	   1076068 ns/op

on my local pc, and I'm even hitting the file system in places.  I don't think it matters.  
Think it's a waste of time to benchmark this  Our db latency is multiple levels of magnitude larger than this, so it doesn't really make much sense to care about this.  .  

Patch tests:
BenchmarkGoTemplateApplier-12    	    3000	    620792 ns/op
BenchmarkPatchTemplateApplier-12    	   10000	    137668 ns/op

### DIFF
--- a/pkg/build/inline_benchmark_test.go
+++ b/pkg/build/inline_benchmark_test.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC
+// Copyright 2019 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/build/inline_benchmark_test.go
+++ b/pkg/build/inline_benchmark_test.go
@@ -1,0 +1,38 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+import (
+	"context"
+	"io/ioutil"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/converter"
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/validate"
+)
+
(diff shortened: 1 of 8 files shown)


pkg/build/inline_benchmark_test.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2018 Google LLC
```
Kashomon: 2019

pkg/build/patchbuild_benchmark_test.go
```diff
@@ -0,0 +1,53 @@
...
+    template: |
+      kind: Pod
+      metadata:
+        namespace: {{.Namespace}}
+        name: {{.PodName}}`
+
+	for i := 0; i < t.N; i++ {
+		c, err := converter.FromYAMLString(component).ToComponent()
+		if err != nil {
+			panic("error parsing component")
```
Kashomon: t.Fatal is a little better. Here and throughout

Kashomon:
Can you also make benchmarks for options/gotmpl and option/patchtmpl? Those are the libraries I'm worried about.

pkg/build/inline_benchmark_test.go
Kashomon: looks wrong still

pkg/build/inline_benchmark_test.go
```diff
@@ -0,0 +1,51 @@
...
+			t.Fatal("error converting component")
+		}
+		inliner := NewLocalInliner("../../examples/component/")
+		component, err := inliner.ComponentFiles(context.Background(), cb, dataPath)
+		if err != nil {
+			t.Fatal("error inlining components")
+		}
+		_, err = converter.FromObject(component).ToYAML()
+		if err != nil {
+			t.Fatal("error converting component")
```
Kashomon: Nit: just t.Fatal(err)

pkg/build/inline_benchmark_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	for i := 0; i < t.N; i++ {
+		cb, err := converter.FromYAML(b).ToComponentBuilder()
+		if err != nil {
+			t.Fatal("error converting component")
+		}
+		inliner := NewLocalInliner("../../examples/component/")
+		component, err := inliner.ComponentFiles(context.Background(), cb, dataPath)
+		if err != nil {
+			t.Fatal("error inlining components")
```
Kashomon: here too: just t.Fatal(err)

pkg/build/inline_benchmark_test.go
Kashomon: Here and through out this PR

Kashomon:
Just a couple of nits. Otherwise LGTM.
/lgtm

Issue #276, opened 2019-05-11
AnnotationMatcher: empty list matches any


Kashomon:
An idea I had with regards to the AnnotationMatcher: An empty list should imply that you can match any annotation value.

PR #277, opened 2019-05-11
Two cleanups for dependency resolution:

Kashomon:
1.  Don't return an error from ComponentVersions. Just return an empty list.
    This is a cleaner / more standard API.
2.  For annotation matching, if the list of values is empty, match anything.
    
Fixes: #276.

### DIFF
--- a/pkg/deps/annotation_matcher.go
+++ b/pkg/deps/annotation_matcher.go
@@ -49,6 +49,9 @@ type AnnotationCriteria struct {
 	// If there are multiple keys (annotations) specified, then all annotations
 	// must match.  In other words, this is a logical AND operation of the passed
 	// in Annotation and the component Annotation.
+	//
+	// If the list of values is empty, then the annotation is matched for any
+	// annotation value.
 	Match map[string][]string
 
 	// Exclude are component annotations that must not be present.  This is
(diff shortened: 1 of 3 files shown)


pkg/deps/annotation_matcher.go
```diff
@@ -75,9 +81,12 @@ func AnnotationMatcher(criteria *AnnotationCriteria) Matcher {
 		}
 
 		matchesAnnot := func(key string, vals []string, annots map[string]string) bool {
+			if _, ok := annots[key]; ok && len(vals) == 0 {
+				return true
```
BradfordMedeiros: It's not wrong, but it's surprising that you would return true given a null set to match against.

pkg/deps/annotation_matcher.go
```diff
@@ -75,9 +81,12 @@ func AnnotationMatcher(criteria *AnnotationCriteria) Matcher {
...
+			}
 			matchesAny := false
 			for _, val := range vals {
-				if m.Annotations[key] == val {
+				if annots[key] == val {
 					matchesAny = true
```
BradfordMedeiros: you return early in the above statement, but here you set a value and return at the end.  One way or the other better.

pkg/deps/annotation_matcher.go
Kashomon: True. But an empty list that always returned false would be pretty useless here

pkg/deps/annotation_matcher.go
Kashomon: Done

PR #281, opened 2019-05-14
Fix edge case: matching doesn't work for 'fixed' components.

Kashomon:


pkg/deps/resolve_test.go
```diff
@@ -761,6 +761,21 @@ spec:
 			},
 			expErrSubstr: "no latest version",
 		},
+		{
+			desc:     "annotations: bad fixed version",
+			universe: annotationSet,
+			comps: []bundle.ComponentReference{
+				{ComponentName: "ann", Version: "1.1.0"},
```
mamaso: This doesn't clearly express that component 'ann' is annotated with 'bad-component'. Can you change naming to make that more explicit or create a new component?

Kashomon: Done

Issue #279, opened 2019-05-14
Should be able to use relative paths on relative paths


justinsb:
https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/eec9db1f2a6d9faac2c864aa50cfe02d35b2935f/pkg/build/path_rewriter.go#L39-L41

When I have a component builder that I reference with a relative path, I would expect to be able to use relative paths in that builder.  But because of the above logic, I can't; I instead must specify the path include the path to the builder.

I expect we should compute an absolute path for the component asap and then use that, or we should allow MakeAbs to work with relative-paths-on-relative-paths.

PR #284, opened 2019-05-28
Some fixes to relative path handling

### LINKED ISSUE #274: componentsuite test fails with nil ref error trying to parse relative paths
When trying to test component builder with relative paths, 
```yaml
apiVersion: bundle.gke.io/v1alpha1
kind: ComponentBuilder
componentName: my-component
version: 0.0.1
appVersion: 0.0.1
objectFiles:
- url: dir/file.yaml
```

Inliner.objectFiles() fails [here](https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/build/inline.go#L204), then
nil ref is that error is ignored [here](https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/testutil/componentsuite/componentsuite.go#L76):
```
comp, err = inliner.ComponentFiles(ctx, bw.ComponentBuilder(), ts.ComponentFile)
		if err != nil {
		}
```
CC @Kashomon

Kashomon:
Now, in bundleio, make the component / bundle files absolute as soon as
possible and return an error in the inliner if they are not absolute.

Also: Fix relative path issues in component testsuite

Fixes:#279
Fixes #274

### DIFF
--- a/pkg/build/inline.go
+++ b/pkg/build/inline.go
@@ -61,7 +61,7 @@ func NewInlinerWithScheme(scheme files.URLScheme, objReader files.FileObjReader)
 	}
 }
 
-// BundleFiles converts dereferences file-references in for bundle files. If
+// BundleFiles inlines file-references in for bundle files. If
 // the bundlePath is defined and not absolute and the scheme is file based
 // scheme, then the path is made absolute before proceeding.
 func (n *Inliner) BundleFiles(ctx context.Context, data *bundle.BundleBuilder, bundlePath string) (*bundle.Bundle, error) {
@@ -224,7 +224,7 @@ func (n *Inliner) objectFiles(ctx context.Context, objFiles []bundle.File, ref b
 
 		contents, err := n.readFile(ctx, cf)
 		if err != nil {
-			return nil, nil, fmt.Errorf("reading file %v for component %v: %v", cf, ref, err)
+			return nil, nil, fmt.Errorf("error reading file %v for component %v: %v", cf, ref, err)
 		}
 		ext := filepath.Ext(cf.URL)
 		if ext == ".yaml" && multiDoc.Match(contents) {
--- a/pkg/build/inline.go
+++ b/pkg/build/inline.go
@@ -36,6 +36,7 @@ type Inliner struct {
 	// Readers reads from the local filesystem.
 	Readers map[files.URLScheme]files.FileObjReader
 
+	// PathRewriter rewrites paths from relative to absolute.
 	PathRewriter *RelativePathRewriter
 }
 
@@ -64,11 +65,15 @@ func NewInlinerWithScheme(scheme files.URLScheme, objReader files.FileObjReader)
 	}
 }
 
-// BundleFiles converts dereferences file-references in for bundle files.
+// BundleFiles converts dereferences file-references in for bundle files. The
+// bundlePath must be an absolute path path this point.
 func (n *Inliner) BundleFiles(ctx context.Context, data *bundle.BundleBuilder, bundlePath string) (*bundle.Bundle, error) {
 	bundleURL, err := url.Parse(bundlePath)
 	if err != nil {
 		return nil, err
+	}
+	if !filepath.IsAbs(bundleURL.Path) {
+		return nil, fmt.Errorf("bundlePath must be absolute but was %s", bundleURL.Path)
 	}
 	var comps []*bundle.Component
 	for _, f := range data.ComponentFiles {
@@ -130,12 +135,16 @@ var onlyWhitespace = regexp.MustCompile(`^\s*$`)
 var multiDoc = regexp.MustCompile("(^|\n)---")
 var nonDNS = regexp.MustCompile(`[^-a-z0-9\.]`)
 
-// ComponentFiles reads file-references for component builder objects.
-// The returned components are copies with the file-references removed.
+// ComponentFiles reads file-references for component builder objects.  The
+// returned components are copies with the file-references removed. The
+// componentPath must be an absolute path at this point.
 func (n *Inliner) ComponentFiles(ctx context.Context, comp *bundle.ComponentBuilder, componentPath string) (*bundle.Component, error) {
 	componentURL, err := url.Parse(componentPath)
 	if err != nil {
 		return nil, err
+	}
+	if !filepath.IsAbs(componentURL.Path) {
+		return nil, fmt.Errorf("componentPath must be absolute but was %s", componentURL.Path)
 	}
 
 	newObjs, tmplBuilders, err := n.objectFiles(ctx, comp.ObjectFiles, comp.ComponentReference(), componentURL)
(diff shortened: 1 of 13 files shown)


pkg/build/inline.go
```diff
@@ -64,19 +65,29 @@ func NewInlinerWithScheme(scheme files.URLScheme, objReader files.FileObjReader)
...
-// BundleFiles converts dereferences file-references in for bundle files.
+// BundleFiles converts dereferences file-references in for bundle files. If
+// the bundlePath is defined and not absolute and the scheme is file based
+// scheme, then the path is made absolute before proceeding.
 func (n *Inliner) BundleFiles(ctx context.Context, data *bundle.BundleBuilder, bundlePath string) (*bundle.Bundle, error) {
 	bundleURL, err := url.Parse(bundlePath)
 	if err != nil {
 		return nil, err
 	}
+	bundleURL, err = n.PathRewriter.Abs(bundleURL)
```
BradfordMedeiros: why is path rewriter a struct member as opposed to just a plain old function?

pkg/build/path_rewriter.go
```diff
@@ -22,23 +22,20 @@ import (
 // RelativePathRewriter rewrites paths when the paths are relative paths.
 type RelativePathRewriter struct{}
 
-// MakeAbs rewrites file paths if the path is relative, from the
+// AbsWithParent rewrites file paths if the path is relative, from the
```
BradfordMedeiros: This whole file should just be plain vanilla functions.  You then have to instance this and struct shove.  That can be avoided.

pkg/build/path_rewriter.go
```diff
@@ -49,5 +46,23 @@ func (rw *RelativePathRewriter) MakeAbs(parent, obj *url.URL) *url.URL {
 	}
 }
 
+// Abs makes an absolute url for URL that has an empty or file-based
+// scheme.
+func (rw *RelativePathRewriter) Abs(obj *url.URL) (*url.URL, error) {
+	npath := obj.Path
+	if (obj.Scheme == "file" || obj.Scheme == "") && !filepath.IsAbs(obj.Path) {
```
BradfordMedeiros: consider obj.Scheme = " dlfjkjasd;lkfjas"

This function returns nil error.

pkg/build/path_rewriter.go
```diff
@@ -22,23 +22,20 @@ import (
...
 		return obj
 	}
 	if parent.Scheme != "file" && parent.Scheme != "" {
 		// Only file schemes are supported.
 		return obj
 	}
-	if !filepath.IsAbs(parent.Path) {
-		return obj
-	}
 	if filepath.IsAbs(obj.Path) {
```
BradfordMedeiros: Function has behavior of returning the objpath is it's already absolute, disregarding the parent argument.

pkg/build/path_rewriter.go
```diff
@@ -49,5 +46,23 @@ func (rw *RelativePathRewriter) MakeAbs(parent, obj *url.URL) *url.URL {
 	}
 }
 
+// Abs makes an absolute url for URL that has an empty or file-based
+// scheme.
+func (rw *RelativePathRewriter) Abs(obj *url.URL) (*url.URL, error) {
```
BradfordMedeiros: pass by value as a default would be nicer so I can tell there's no mutation going on.

Kashomon: you mean url.URL?

Technically go is always pass by value since references are themselves copied. https://dave.cheney.net/2017/04/29/there-is-no-pass-by-reference-in-go

pkg/build/path_rewriter.go
Kashomon: That seems fine to me. This doesn't care about unknown schemes. But, I would think that URL parsing would fail earlier

BradfordMedeiros: this function needs to work on its own, you shouldn't depend on other functions behavior.  just rewrite it so the unknown defaults to an error instead of nil error.

pkg/build/path_rewriter.go
BradfordMedeiros: obj is a pointer.   It looks like you're not mutating obj. 
It would be clearer to not pass a pointer, so I know you're not mutating the value of obj. 

Not a big deal.

pkg/build/inline.go
Kashomon: Made a POF

pkg/build/path_rewriter.go
Kashomon: Done

pkg/build/path_rewriter.go
Kashomon: Ack

pkg/build/path_rewriter.go
Kashomon: Yes, that seems like the right behavior. If the path is already absolute, then it doesn't need to be made absolute

pkg/build/path_rewriter.go
Kashomon: Not following -- this makes it much cleaner. Here, Abs ignores all schemes it doesn't know about. I can add the checks here or in the parent file, but it's not clear to me which is better -- in the parent it adds duplication. 

I've changed the name to makeAbsForFileScheme to be clearer.

pkg/build/inline.go
```diff
@@ -59,24 +57,32 @@ func NewInlinerWithScheme(scheme files.URLScheme, objReader files.FileObjReader)
...
 	}
 	return &Inliner{
-		Readers:      rdrMap,
-		PathRewriter: DefaultPathRewriter,
+		Readers: rdrMap,
 	}
 }
 
-// BundleFiles converts dereferences file-references in for bundle files.
+// BundleFiles converts dereferences file-references in for bundle files. If
```
justinsb: Not really sure what "converts dereferences" means - typo?

pkg/build/inline.go
```diff
@@ -205,11 +220,11 @@ func (n *Inliner) objectFiles(ctx context.Context, objFiles []bundle.File, ref b
 		if err != nil {
 			return nil, nil, err
 		}
-		cf.URL = n.PathRewriter.MakeAbs(componentPath, furl).String()
+		cf.URL = makeAbsWithParent(componentPath, furl).String()
 
 		contents, err := n.readFile(ctx, cf)
 		if err != nil {
-			return nil, nil, fmt.Errorf("error reading file %v for component %v: %v", cf, ref, err)
+			return nil, nil, fmt.Errorf("reading file %v for component %v: %v", cf, ref, err)
```
justinsb: I think the old form (with "error reading...") was more go.

justinsb:
Two non-code nits, but otherwise LGTM.  Going to pull it down & test it out (but if you've already done so, feel free to merge when you're happy)!

pkg/build/inline.go
Kashomon: Done

pkg/build/inline.go
Kashomon: Done

PR #285, opened 2019-06-05
Add support for already-inlined objects in ComponentBuilder

Kashomon:
This PR adds support for component files + components in the component builders. This is nice for just inlining small objects that don't change often. In particular, this is useful for Requirements objects,  ObjectTemplateBuilders, and small PatchTemplates

### DIFF
--- a/pkg/build/inline.go
+++ b/pkg/build/inline.go
@@ -158,6 +158,12 @@ func (n *Inliner) ComponentFiles(ctx context.Context, comp *bundle.ComponentBuil
 		return nil, err
 	}
 
+	objs, tbuilders := n.separateObjects(comp.Objects, componentURL)
+	newObjs = append(newObjs, objs...)
+	for k, vals := range tbuilders {
+		tmplBuilders[k] = append(tmplBuilders[k], vals...)
+	}
+
 	tmplObjs, err := n.objectTemplateBuilders(ctx, tmplBuilders, comp.ComponentReference())
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/build/inline.go
```diff
@@ -237,6 +243,10 @@ func (n *Inliner) objectFiles(ctx context.Context, objFiles []bundle.File, ref b
 				if err != nil {
 					return nil, nil, fmt.Errorf("converting multi-doc object number %d for component %v, %v", i, ref, err)
 				}
+				// It's easiest to extract the ObjectTemplateBuilder objects here,
+				// because, if relative paths are used within the
+				// ObjectTemplateBuilder, they'll be relative to the path used for the
+				// ObjectTemplateBuilder
```
mikedanese: Side: we should figure out how to preserve location information through the expansion process.

pkg/build/path_rewriter.go
```diff
@@ -33,10 +33,15 @@ func makeAbsWithParent(parent, obj *url.URL) *url.URL {
 	if path.IsAbs(obj.Path) {
 		return obj
 	}
+	parentDir := path.Dir(parent.Path)
```
mikedanese: If parent is the path of a component, why would it every be a directory?

pkg/build/inline.go
Kashomon: Oh probably. I used to have it and thought it was unnecessary, so I removed it. I could preserve it as an annotation.

pkg/build/path_rewriter.go
Kashomon: Maybe I'm not following.

parent.Path is the full path of the component. This just grabs the directory. `foo/bar/biff-component.yaml => foo/bar`

mikedanese: I'm wondering where we would ever hit the if below this. i.e. why wasn't this correct before?

Kashomon: Man, I can't remember. this pr is almost two months old. XD

PR #288, opened 2019-06-23
Enable object deserialization test in component testsuite.

Kashomon:
This PR modifies the component test-suite to have a canKubeDeserialize
field that checks whether objects can be deserialized using the internal
Kubernetes deserialization logic. This relies on the schema in the
patchtmpl (which perhaps should be put in a more common place, like
`pkg/options`)

A couple smaller changes:

- disable jsonnet importing example for now
- add jsonnet to the multi-applier
- add a deserialize test case to all examples
- add a test case for jsonnet

### DIFF
--- a/examples/jsonnet/bazel_init_test.go
+++ b/examples/jsonnet/bazel_init_test.go
@@ -1,0 +1,23 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package jsonnet_test
+
+import (
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/testutil"
+)
+
+func init() {
+	testutil.ChangeToBazelDir("examples/jsonnet")
+}
--- a/pkg/testutil/componentsuite/BUILD.bazel
+++ b/pkg/testutil/componentsuite/BUILD.bazel
@@ -2,6 +2,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
+    testonly = True,
     srcs = [
         "componentsuite.go",
         "schema.go",
@@ -27,6 +28,7 @@ go_library(
 
 go_test(
     name = "go_default_test",
+    testonly = True,
     srcs = [
         "bazel_init_test.go",
         "componentsuite_test.go",
--- a/pkg/testutil/componentsuite/BUILD.bazel
+++ b/pkg/testutil/componentsuite/BUILD.bazel
@@ -5,6 +5,7 @@ go_library(
     srcs = [
         "componentsuite.go",
         "schema.go",
+        "validate.go",
     ],
     importpath = "github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/testutil/componentsuite",
     visibility = ["//visibility:public"],
@@ -14,10 +15,13 @@ go_library(
         "//pkg/converter:go_default_library",
         "//pkg/filter:go_default_library",
         "//pkg/options:go_default_library",
+        "//pkg/options/multi:go_default_library",
         "//pkg/options/patchtmpl:go_default_library",
         "//pkg/testutil:go_default_library",
         "//pkg/validate:go_default_library",
         "//pkg/wrapper:go_default_library",
+        "//vendor/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured:go_default_library",
+        "//vendor/k8s.io/apimachinery/pkg/runtime:go_default_library",
     ],
 )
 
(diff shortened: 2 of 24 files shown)


examples/jsonnet/bazel_init_test.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package jsonnet_test
+
+import (
+	"github.com/GoogleCloudPlatform/k8s-cluster-bundle/pkg/testutil"
+)
+
+func init() {
+	testutil.ChangeToBazelDir("examples/jsonnet")
```
mikedanese: This is the default testdir already..?

pkg/testutil/componentsuite/componentsuite.go
```diff
@@ -130,88 +127,20 @@ func runApply(t *testing.T, comp *bundle.Component, tc *TestCase) *bundle.Compon
 	applyOpts := &filter.Options{}
 	applyOpts.Annotations = tc.Apply.Filters
```
mikedanese: Unused?

pkg/testutil/componentsuite/componentsuite.go
```diff
@@ -130,88 +127,20 @@ func runApply(t *testing.T, comp *bundle.Component, tc *TestCase) *bundle.Compon
...
 	comp, err := applier.ApplyOptions(comp, tc.Apply.Options)
 	cerr := testutil.CheckErrorCases(err, tc.Expect.ApplyErrSubstr)
 	if cerr != nil {
 		t.Fatal(cerr)
 	}
+
+	o, _ := converter.FromObject(comp).ToYAMLString()
+	t.Logf("obj: %v", o)
+
 	if err != nil {
```
mikedanese: Why isn't this checked after applier.ApplyOptions?

pkg/testutil/componentsuite/componentsuite.go
```diff
@@ -130,88 +127,20 @@ func runApply(t *testing.T, comp *bundle.Component, tc *TestCase) *bundle.Compon
...
-		true /* includeTemplates, for debugging */)
+	applier := multi.NewDefaultApplier()
 
 	comp, err := applier.ApplyOptions(comp, tc.Apply.Options)
 	cerr := testutil.CheckErrorCases(err, tc.Expect.ApplyErrSubstr)
 	if cerr != nil {
 		t.Fatal(cerr)
 	}
+
+	o, _ := converter.FromObject(comp).ToYAMLString()
```
mikedanese: Check the error.

pkg/testutil/componentsuite/BUILD.bazel
```diff
@@ -5,6 +5,7 @@ go_library(
     srcs = [
         "componentsuite.go",
         "schema.go",
+        "validate.go",
```
mikedanese: Should this library have the `testonly = True` attribute?

pkg/testutil/componentsuite/validate.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, obj := range comp.Spec.Objects {
+		key := objKeyFromObj(obj)
+		objByt, err := converter.FromObject(obj).ToJSON()
+		if err != nil {
+			// This would be pretty unlikely
+			t.Errorf("Error converting object %v to JSON: %v", key, err)
+			continue
+		}
+
+		_, err = runtime.Decode(deserializer, objByt)
```
mikedanese: if _, err := ...; err != nil

pkg/testutil/componentsuite/validate.go
```diff
@@ -0,0 +1,137 @@
...
+
+	checkObjectProperties(t, comp, tc)
+}
+
+func checkObjectsDeserialize(t *testing.T, comp *bundle.Component) {
+	scheme := patchtmpl.DefaultPatcherScheme()
+	deserializer := scheme.Codecs.UniversalDeserializer()
+
+	for _, obj := range comp.Spec.Objects {
+		key := objKeyFromObj(obj)
```
mikedanese: Wrap in a t.Run named with key?

pkg/testutil/componentsuite/validate.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	for key := range objCheckMap {
+		if _, ok := objMap[key]; !ok {
+			t.Errorf("Got object-keys %q, but expected to find object %v", stringMapKeys(objMap), key)
+		}
+	}
+}
+
+func stringMapKeys(m map[objKey]string) []string {
```
mikedanese: You can use sets.StringKeySet.

examples/jsonnet/bazel_init_test.go
Kashomon: no, unfortunately. The default testdir is the workspace root.

mikedanese: > These files are available using the relative path path/to/data/file. In tests, it is also possible to refer to them by joining the paths of the test's source directory and the workspace-relative path, e.g. ${TEST_SRCDIR}/workspace/path/to/data/file.

From https://docs.bazel.build/versions/master/build-ref.html#data

"relative" indicates that the working dir is configured properly. Do the tests pass if you remove this file?

Kashomon: No. The tricky thing is I want the tests to pass both with Bazel and with normal `go test`. So, the solution here is what's used in `third_party` (`third_party_go_test`) have an explicit change-dir that's executed only for bazel paths so `ioutil` works for both `go test` and  `bazel test`

pkg/testutil/componentsuite/componentsuite.go
Kashomon: Removed. nice catch. Although it exposes an inconsistency in the schema -- removed that too.

pkg/testutil/componentsuite/validate.go
Kashomon: /shrug. I think it's a little heavy weight to rely on K8S ApiMachinery here personally.

PR #290, opened 2019-10-25
Preserve PatchType when creating PatchTemplate from builder

garron:
In trying to use JSONPatch with a PatchTemplateBuilder, I noticed that the PatchType field was not getting carried through to the created PatchTemplate. This change addresses that.

I also added a testcase (which previously failed).


Kashomon:
Awesome! Much appreciated.

PR #291, opened 2020-02-05
gotmpl Applier: Add text/template options config

ahmedtd:
This commit lets the caller specify text/template options when creating a gotmpl applier.  The only meaningful option to pass is `missingkey={ignore,error,zero}`, to control the behavior when the template references an unset variable.


Kashomon:
Nit: If you wouldn't mind, it would be nice to have consistent options methods between the other two options appliers (I deleted jsonnet). Otherwise, happy to submit this as is.

/lgtm

ahmedtd:
I added a similar config setup to patchtmpl.  multi seems a bit too simple for this setup, and I can't think of any options it would want to support in the future.

PR #299, opened 2020-06-22
Add storage/v1beta1 to default patcher

misterikkit:
This change allows clients to patch CSIDriver and CSINode objects in
cluster bundles. These types were added to v1beta1 in kubernetes 1.14.0.

Tested: bazel test //...


misterikkit:
@Kashomon @avrittrohwer You seem like the right people to review this.

Kashomon:
/lgtm

Looks like prow isn't running again for some reason, but I can't imagine this will break things

PR #300, opened 2020-06-23
Update the 0.12 release branch

Kashomon:
Update the release branch so we can do a 0.12.3 release.

+ update the version

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -16,4 +16,4 @@
 package version
 
 // BundlectlVersion is the version used by bundlectl.
-const BundlectlVersion = "0.12.0"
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -16,4 +16,4 @@
 package version
 
 // BundlectlVersion is the version used by bundlectl.
-const BundlectlVersion = "0.12.0"
+const BundlectlVersion = "0.12.3"
```
avrittrohwer, line 19: Is there a reason this is being bumped to .3 instead of .1?

Kashomon: Uhh. because I forgot to do it the previous 2 releases

PR #302, opened 2020-07-15
Support optional filtering functions in the image extractor.

JonathanSun:


pkg/find/images.go
```diff
@@ -65,12 +71,18 @@ func (b *ImageFinder) AllContainerImages() []*ContainerImage {
 
 // ContainerImages returns all the images from a single Kubernetes object.
 func (b *ImageFinder) ContainerImages(key bundle.ComponentReference, st *unstructured.Unstructured) []*ContainerImage {
+	return b.FilteredContainerImages(nil, key, st)
+}
+
+// FilteredContainerImages returns all the images from a single Kubernetes object,
+// filtered by the specified function.
+func (b *ImageFinder) FilteredContainerImages(filter func(fieldName string, parentFieldName string, img string) bool, key bundle.ComponentReference, st *unstructured.Unstructured) []*ContainerImage {
```
Kashomon, line 79: What's the intent of the filter function? 

Maybe we can make a separate type and document it?

JonathanSun: Done.

JonathanSun:
PTAL

PR #303, opened 2020-08-04
Add webhook v1beta1 to default patcher

wangzhen127:
This change allows clients to patch webhook objects in cluster bundles. `admissionregistration.k8s.io/v1beta1` is available since kubernetes 1.9


wangzhen127:
@Kashomon @avrittrohwer You seem like the right people to review this.

PR #306, opened 2020-08-04
Add a TemplateBuilder example.

Kashomon:
This PR adds a new (go) template builder example to the `examples/` directory.

### DIFF
--- a/examples/templatebuilder/etcd-template.yaml
+++ b/examples/templatebuilder/etcd-template.yaml
@@ -1,0 +1,81 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: etcd-server
+  namespace: "{{.Namespace}}"
+spec:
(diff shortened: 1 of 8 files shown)


examples/templatebuilder/etcd-template.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: v1
```
avrittrohwer, line 1: nit: make this file as simple as it can be so that the intent of the example is clearer.

Kashomon: Probably followup to do this since all the examples use `etcd`

Kashomon: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/307

Issue #305, opened 2020-08-04
Add Template to ComponentBuilder


Kashomon:
Today, the ComponentBuilder looks something like:

https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/examples/component/etcd-component-builder.yaml

```yaml
apiVersion: bundle.gke.io/v1alpha1
kind: ComponentBuilder
componentName: etcd-component
version: 30.0.2
objectFiles:
- url: template-builder.yaml
```

and then the template-builder will reference via

```yaml
apiVersion: bundle.gke.io/v1alpha1
kind: ObjectTemplateBuilder
file:
- url: etcd-template.yaml
```

This is a silly amount of boiler plate. It would be nice to reduce this into the component builder:

```yaml
apiVersion: bundle.gke.io/v1alpha1
kind: ComponentBuilder
componentName: etcd-component
version: 30.0.2
objectFiles:
- url: some-non-templated-file.yaml
templateFiles:
- url: etcd-template
```

Much cleaner

Kashomon:
This would require:

- Changing the builder type: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/apis/bundle/v1alpha1/builder_types.go. We would probably *not* want to add the open API schema for now.

- Changing the inlining logic: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/build/inline.go#L143

nicolexin:
@Kashomon Do we want to maintain backwards compatibility?

PR #308, opened 2020-08-11
Add TemplateFiles to ComponentBuilder

nicolexin:
Fixes: https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/issues/305

### DIFF
--- a/pkg/apis/bundle/v1alpha1/builder_types.go
+++ b/pkg/apis/bundle/v1alpha1/builder_types.go
@@ -71,25 +71,32 @@ type ComponentBuilder struct {
 	// it should be representable as either YAML or JSON.
 	ObjectFiles []File `json:"objectFiles,omitempty"`
 
-	// Templates that are specified via a File-URL. The process of inlining a
-	// component turns template files into objects with template.  During the
-	// inline process, if the file is YAML-formatted and contains multiple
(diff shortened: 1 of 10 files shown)


pkg/apis/bundle/v1alpha1/builder_types.go
```diff
@@ -71,6 +71,20 @@ type ComponentBuilder struct {
...
 
+	// Templates that are specified via a File-URL. The process of inlining a
+	// component turns template files into objects with template.  During the
+	// inline process, if the file is YAML-formatted and contains multiple
+	// objects in the YAML-doc, the objects will be split into separate inline
+	// objects. In other words, one object file may result in multiple objects.
+	//
+	// Each object file must be parsable into a Struct: In other words,
+	// it should be representable as either YAML or JSON.
+	TemplateFiles []File `json:"templateFiles,omitempty"`
```
Kashomon, line 82: I was imagining something like

```
TemplateFiles: 
- templateType: foo
   url: bar.yaml
```

Although that doesn't quite work with File type. So maybe instead

```
TemplateFiles: 
- templateType: foo
   files: 
   - url: bar.yaml
```

with 

```
type TemplateFile struct {
 TemplateType TemplateType
  Files []File
}
```

This would allow the possibility of specifying multiple different template types in on Component, which I rather like for flexibility

PR #309, opened 2020-09-01
filter.go: add generic filtering mechanism

foolusion:
This adds a Select function that takes a slice of components and a
ComponentPredicate function. This allows both AND and OR comparisons.


Kashomon:
Needs some tests, but I like the idea. Is the plan to eventually remove the old filtering logic?

pkg/filter/filter.go
```diff
@@ -297,3 +297,178 @@ func matches(d *objectData, o *Options) bool {
...
+				}
+			}
+		}
+		return false
+	}
+}
+
+// UnstructuredNamespaceIn checks if any of the components object.GetNamespace()
+// are in namespaces.
+func UnstructuredNamespaceIn(namespaces ...string) ComponentPredicate {
```
Kashomon, line 334: Maybe "ObjectNamespaceIn" in components, the unstructured resources are generally referred to as Objects.

foolusion:
> Needs some tests, but I like the idea. Is the plan to eventually remove the old filtering logic?

I don't know, we could convert the internal logic to use this, but replacing it would break API compatibility. I think we also need to write a similar SelectObjects method that returns a list of objects.

pkg/filter/filter.go
foolusion: Done

pkg/filter/filter.go
```diff
@@ -297,3 +297,202 @@ func matches(d *objectData, o *Options) bool {
...
+				return true
+			}
+		}
+		return false
+	}
+}
+
+// And returns true if every ComponentPredicate function returns true and
+// returns false otherwise.
+func And(componentPredicates ...ComponentPredicate) ComponentPredicate {
```
Kashomon, line 453: Do you need a Not method? Otherwise maybe add a Filter method?

pkg/filter/example_select_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+func toComponentList(components []*bundle.Component) []component {
+	out := make([]component, len(components))
+	for i, c := range components {
+		out[i] = toComponent(c)
+	}
+	return out
+}
+
+func Example_select() {
```
Kashomon, line 116: This isn't a test method...?

Also, I would recommend a table test that has more extensive cases.

pkg/filter/BUILD.bazel
```diff
@@ -15,7 +15,11 @@ go_library(
 
 go_test(
     name = "go_default_test",
-    srcs = ["filter_test.go"],
+    srcs = [
+        "example_selectObjects_test.go",
```
Kashomon, line 19: not seeing this file.

pkg/filter/example_selectObjects_test.go
```diff
@@ -0,0 +1,36 @@
+package filter_test
```
Kashomon, line 1: file comment: please make all files snake_case.go

pkg/filter/example_select_test.go
Kashomon: Oh, I see, it's an example. I've honestly never used examples.

Still, some table tests might be nice.

Kashomon:
/lgtm

I'll leave it to you to merge if you feel you want to do more polishing.

Kashomon (approved):

PR #310, opened 2020-10-07
Support invert match in the internal function `matches`.

Random-Liu:
The invert match should be implemented in the underlying `matches` function, or else functions like `MatchesComponent` and `MatchesObject` won't have the support.

And please note that they are used by patch.go https://github.com/GoogleCloudPlatform/k8s-cluster-bundle/blob/master/pkg/options/patchtmpl/patch.go#L359


avrittrohwer (approved):
Looks good to me

PR #311, opened 2020-10-09
[Cherrypick] Support invert match in the internal function `matches`.

Random-Liu:
Cherrypick the fix into release-0.12


Random-Liu:
@Kashomon Not sure whether we should get this cherrypicked and cut a 0.12.6 release, or we should just reference the commit. Do you have any suggestion?

PR #318, opened 2022-05-02
Modify examples/patchbuilder/deployment-patch-builder.yaml to reproduce number comparison issue.

jiayingz:
With the change, running `bazel test //examples/patchbuilder/...` failed with:
    --- FAIL: TestComponentSuite/Deserialize_Objects (0.00s)
        componentsuite.go:117: got error "cannot build PatchTemplate from PatchTemplateBuilder \"\": error executing template: template: ptb:12:15: executing \"ptb\" at <ge .Port 1>: error calling ge: incompatible types for comparison" but expected no error


artemispax (approved):
Let's merge and then I'll find a fix

PR #319, opened 2022-05-10
Add test cases for patch template builder edgcase

artemispax:
Here, I added test cases for patch template builder edgecase.

There are two issues:

1. Mixing go-templates and patch template builders don't work terribly
   well. Because the patch template builder has built into a secondary
   template, any complex go-template properties need to be applied in
   the build phase.
2. The YAML/JSON conversion schema can't tell the difference between an
   integer and a float.

These are unfortunately limitations of the patch template design.

Here, I add a possible solution (put the go-template stuff in
patch-building).

An alternative is to use ObjectTemplateBuilders.

### DIFF
--- a/examples/patchbuilder/deployment-patch-builder.yaml
+++ b/examples/patchbuilder/deployment-patch-builder.yaml
@@ -7,6 +7,14 @@ buildSchema:
     ImageTag:
       description: The image tag to deploy.
       type: string
+    PortStartGuard:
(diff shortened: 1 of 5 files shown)


mtaufen:
I can neither confirm nor deny that this PR is a good idea.

But I can LGTM.

/lgtm

mtaufen (approved):

PR #325, opened 2022-11-16
Add safetext YAML processing

artemispax:
This PR Adds an optional safetext yaml Library library that can be opted into using an option in the patch template builder and also the object templates.

Also: updates to go 1.19


pkg/apis/bundle/v1alpha1/builder_types.go
```diff
@@ -92,6 +92,9 @@ type TemplateFileSet struct {
 	// to Go Templates during build if left unspecified.
 	TemplateType TemplateType `json:"templateType,omitempty"`
 
+	// UseSafeYAMLTemplater determines whether to use SafeYAMLTemplating.
+	UseSafeYAMLTemplater *bool `json:"useSafeYamlTemplater,omitempty"`
```
mtaufen, line 96: Do any of these UseSafeYAMLTemplater options override each other depending on where they're set? If so we should make the overriding/inheritance behavior clear in doc comments/actual docs.

artemispax: Good point. Added some laungage that this is a default and doesn't override the downstream template settings.

mtaufen (approved):

PR #339, opened 2024-07-16
Wrap master address with quotes.

krzykwas:
The existing quotes get removed during template processing, resulting in the IP address being inserted into the yaml verbatim. This works with IPv4 addresses, but causes parsing errors with IPv6 addresses.

### DIFF
--- a/examples/cluster/kubeproxy/kubeproxy.yaml
+++ b/examples/cluster/kubeproxy/kubeproxy.yaml
@@ -43,7 +43,7 @@ spec:
         - kube-proxy {{cluster_cidr}} --resource-container="" --oom-score-adj=-998 {{params}} 1>>/var/log/kube-proxy.log 2>&1
         env:
         - name: KUBERNETES_SERVICE_HOST
-          value: "{{kubernetes_service_host_env_value}}"
+          value: "\"{{kubernetes_service_host_env_value}}\""
         securityContext:
           privileged: true
         volumeMounts:


justinsb (approved):

examples/cluster/kubeproxy/kubeproxy.yaml
```diff
@@ -43,7 +43,7 @@ spec:
         - kube-proxy {{cluster_cidr}} --resource-container="" --oom-score-adj=-998 {{params}} 1>>/var/log/kube-proxy.log 2>&1
         env:
         - name: KUBERNETES_SERVICE_HOST
-          value: "{{kubernetes_service_host_env_value}}"
+          value: "\"{{kubernetes_service_host_env_value}}\""
```
mtaufen, line 46: The env var itself needs to contain the quotes? That seems weird?

mtaufen: If the existing quotes get removed during template processing, does this result in `value: \"10.10.10.10\"` in the rendered template, or `value: "10.10.10.10"`?

krzykwas: On the master node I found something like:

```
- name: KUBERNETES_SERVICE_HOST
- value: fd20:2001:1::
```

Though apparently it's not because of this file, there is rather a different copy that needs a fix instead.

krzykwas: Thanks for your offline investigation @mtaufen!

krzykwas:
This file is apparently not the one that needs the fix, hence I'll close the PR.

